PR #1, opened 2019-08-22
Option, Result, and update Readme

samuelgruetter:
Let's get started with the standard library!

### DIFF
--- a/src/Std.dfy
+++ b/src/Std.dfy
@@ -1,0 +1,52 @@
+module Std {
+
+  newtype uint8 =  x | 0 <= x < 0x100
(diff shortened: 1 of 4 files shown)


src/Std.dfy
```diff
@@ -0,0 +1,52 @@
+module Std {
+
+  newtype uint8 =  x | 0 <= x < 0x100
```
JunyoungLim: Will this `uint8` type be used instead of an unsigned `byte`?

samuelgruetter: Yes, that's what I would suggest, because if you just say `byte`, it's not clear if it's signed or unsigned.

RustanLeino:
Exemplary!

PR #2, opened 2019-09-03
Make `Result` type `:-` friendly

RustanLeino:
Add `IsFailure`, `PropagateFailure`, and `Extract` members to datatype `Result` to make
it usable with `:-` assignments in Dafny.

Update example to use `:-` with `Result`.

### DIFF
--- a/examples/Std/OptionAndResult.dfy
+++ b/examples/Std/OptionAndResult.dfy
@@ -97,32 +97,19 @@ module Demo {
     }
   }
 
-  method TestMyFilesystem() {
+  method TestMyFilesystem() returns (r: Result<()>) {
     var fs := new MyFilesystem();
-    // Note: these verbose "outcome.Failure?" patterns will soon
-    // not be needed any more, see https://github.com/dafny-lang/rfcs/pull/1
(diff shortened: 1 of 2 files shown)


examples/Std/OptionAndResult.dfy
```diff
@@ -97,32 +97,19 @@ module Demo {
...
-    var fileResult: Result<string> := fs.ReadFile("test.txt");
-    if outcome.Failure? {
-      print fileResult.error, "\n";
-      return;
-    }
-    if fileResult.Success? {
-      print fileResult.value, "\n";
-    }
+    var _ :- fs.CreateFile("test.txt");
+    var _ :- fs.WriteFile("test.txt", "Test dummy file");
```
samuelgruetter: When I implemented `:-`, I intended to support statements starting with `:-`, so that you could drop the `var _ :- ` part in the above code. Did you try that? Does it work? Or do you get an error saying that there must not be an `Extract` method? In any case, I think there should still be a TODO note here

RustanLeino:
This has now been added in another PR.

PR #3, opened 2019-09-04
Base64 Encoding Library

JunyoungLim:
This PR introduces the Base64 Library, which has the following key function methods:
```dafny
function method Decode(s: seq<char>): (b: Result<seq<uint8>>)
```
which takes a string, and returns a sequence of unsigned bytes when the string is of Base64 encoding, or a failure otherwise,

and
```dafny
function method Encode(b: seq<uint8>): (s: seq<char>)
    ensures Decode(s) == Success(b)
```
which takes a sequence of unsigned bytes and returns an encoded string with the Base64 scheme, and guarantees that the `Decode` function will be able to successfully decode the output of the `Encode` function.

### DIFF
--- a/examples/Base64/Base64Encode.dfy
+++ b/examples/Base64/Base64Encode.dfy
@@ -1,15 +1,18 @@
 include "../../src/Base64.dfy"
 
 module Demo {
+    import opened Std
+    import opened Base64
+    
     method TestBase64(msg: string, expected: string)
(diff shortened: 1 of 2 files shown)


samuelgruetter:
ping @RustanLeino, could you please review this?

RustanLeino:
This slipped through the cracks for me, I'm afraid. It looks good. My only comment is that we now use 2 spaces for the indent instead of 4.

PR #7, opened 2021-07-27
Dafny Standard Library

mmwinchell:
Hello everyone! For the past couple of months, we have been working on creating a standard library for Dafny. We would love to hear your feedback on our `Collections` and `Nonlinear Arithmetic` libraries. The original discussion about the Dafny standard library can be found [here](https://github.com/dafny-lang/dafny/issues/1236) .


robin-aws:
This is so exciting! Thanks so much for putting it together. I hope you're prepared for a whole lot of good-natured feedback from a lot of opinionated Dafny programmers. :)

I wanted to mention right away that it's possible putting all this code directly in `dafny-lang/dafny` might help maintain it as Dafny evolves. We'll try to figure that out as we review the actual content.

src/NonlinearArithmetic/Internals/DivInternals.dfy
```diff
@@ -0,0 +1,196 @@
...
+*  Modifications and Extensions: Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+************************************************************************************/
+
+/* lemmas and functions in this file are used in the proofs in DivMod.dfy 
+
+Specs/implements mathematical div and mod, not the C version.
+This may produce "surprising" results for negative values
+For example, -3 div 5 is -1 and -3 mod 5 is 2.
+Note this is consistent: -3 * -1 + 2 == 5 */
```
tchajed, line 17: I find this comment strange because I don't know what C's div and mod do (these results aren't surprising to me). Maybe you could mention the consistency property you're alluding to? I think it's `(x div n) * n + (x mod n) == x`, and maybe it would help to mention `0 <= x mod n < n`.

src/NonlinearArithmetic/Mul.dfy
```diff
@@ -0,0 +1,567 @@
...
+
+include "Internals/MulInternalsNonlinear.dfy"
+include "Internals/MulInternals.dfy"
+
+module Mul {
+
+  import MulINL = MulInternalsNonlinear
+  import opened MulInternals
+
+  /* the common syntax of multiplication results in the same product as multiplication
```
tchajed, line 23: Can you write "builtin" instead of "common"? Took me a second to figure out what you meant.

src/NonlinearArithmetic/Mul.dfy
```diff
@@ -0,0 +1,567 @@
...
+    ensures forall x: int, y: int :: x * y == mul_recursive(x, y)
+  {
+    forall x: int, y: int
+      ensures x * y == mul_recursive(x, y)
+    {
+      lemma_mul_is_mul_recursive(x, y);
+    }
+  }
+
+  /* the common syntax of multiplying two positive integers results in the same product as
```
tchajed, line 43: same comment about "common"

src/NonlinearArithmetic/Mul.dfy
```diff
@@ -0,0 +1,567 @@
...
+        :: x * (y * z) == (x * y) * z
+  {
+    forall (x: int, y: int, z: int)
+      ensures x * (y * z) == (x * y) * z
+    {
+      lemma_mul_is_associative(x, y, z);
+    }
+  }
+
+  /* multiplcation is commutative */
```
tchajed, line 118: typo in "multiplcation"

src/NonlinearArithmetic/DivMod.dfy
```diff
@@ -0,0 +1,1569 @@
...
+  {
+    lemma_mul_nonzero_auto();
+    forall x: int, c: nat, d: nat | 0 <= x && 0 < c && 0 < d
+      ensures (x / c) / d == x / (c * d)
+    {
+      lemma_div_denominator(x, c, d);
+    }
+  }
+
+  /* mutltiplying an integer by a fraction is equivalent to multiplying the integer by the
```
tchajed, line 587: typo "mutltiplying"

tchajed (commented):
Very cool! I left a few notes from skimming the changes.

One thing I wonder about is how to make use of the non-linear arithmetic part of the library. It seems challenging to come up with documentation for searching through relevant lemmas. In Coq I rely heavily on the `Search` command to do this, but Dafny doesn't have something like that.

src/Options.dfy
```diff
@@ -0,0 +1,42 @@
...
+    
+    function method unwrap_or(default: T): T 
+    {
+      match this
+      case Some(v) => v
+      case None => default
+    }
+  }
+  
+  datatype Result<T> = | Success(value: T) | Failure(error: string) {
```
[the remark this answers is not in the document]
tchajed: In fact it can implement the entire "failure-compatible type" API; see https://dafny-lang.github.io/dafny//DafnyRef/DafnyRef#1971-failure-compatible-types.

src/Collections/Maps/Maps.dfy
```diff
@@ -0,0 +1,132 @@
...
+
+  /* Keep all key-value pairs corresponding to the set of keys provided. */
+  function method {:opaque} restrict<X, Y>(m: map<X, Y>, xs: set<X>): (m': map<X, Y>)
+    ensures m' == remove_keys(m, m.Keys - xs)
+  {
+    map x | x in xs && x in m :: m[x]
+  }
+
+  /* True iff two maps contain the same key-value pairs for intersecting
+  keys. */
```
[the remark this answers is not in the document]
sarahc7: I changed the predicate description - thanks for catching that!

src/Collections/Maps/Maps.dfy
```diff
@@ -0,0 +1,132 @@
...
+
+include "../../Options.dfy"
+
+module Maps {
+  import opened Options
+
+  function method get<X, Y>(m: map<X, Y>, x: X): Option<Y>
+  {
+	  if x in m then Some(m[x]) else None
+	}
```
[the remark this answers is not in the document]
sarahc7: Fixed.

src/Collections/Maps/Maps.dfy
```diff
@@ -0,0 +1,132 @@
...
+
+  function method {:opaque} to_imap<X, Y>(m: map<X, Y>): (m': imap<X, Y>)
+    ensures forall x {:trigger m'[x]} :: x in m ==> x in m' && m'[x] == m[x]
+    ensures forall x {:trigger x in m'} :: x in m' ==> x in m
+  {
+	  imap x | x in m :: m[x]
+	}
```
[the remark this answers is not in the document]
sarahc7: Fixed.

src/NonlinearArithmetic/Mul.dfy
mmwinchell: Thanks for the wording suggestion!

src/NonlinearArithmetic/DivMod.dfy
mmwinchell: Thanks for catching this

lit.site.cfg
```diff
@@ -0,0 +1,195 @@
+# -*- Python -*-
```
robin-aws, line 1: I get this is here in order to test the library in the same way that Dafny itself is tested, but it's a poor match for an actual Dafny code base since it's only testing that everything verifies, and doesn't need to assert what happens on invalid Dafny files as the Dafny test suite does.

I'd be happier to see at least a few `{:test}` methods that CI can run instead, as demonstrated by this test file: https://github.com/dafny-lang/dafny/blob/master/Test/DafnyTests/TestAttribute.dfy. It's a little unfortunate that the only supported way to run them is via `dotnet test`, but as I just suggested on https://github.com/dafny-lang/dafny/issues/1322#issuecomment-889184653 perhaps we can add a `dafny test` command to make that simpler.

Yes, I know, it probably feels weird that I'm asking for dynamically executed tests on a Dafny library. In my experience it helps to guard against specification errors, though.

src/Options.dfy
```diff
@@ -0,0 +1,42 @@
+// RUN: %dafny "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+/*********************************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+**********************************************************************************************/
+
+module Options {
```
robin-aws, line 9: It feels weird to me to define `Result` in this module, as a kind of "Option".

src/Options.dfy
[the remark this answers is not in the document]
robin-aws: I'd also prefer this use a second type parameter for the error result rather than `string`.

src/Options.dfy
```diff
@@ -0,0 +1,42 @@
...
+  }
+  
+  datatype Result<T> = | Success(value: T) | Failure(error: string) {
+    function method to_option(): Option<T> 
+    {
+      match this
+      case Success(s) => Some(s)
+      case Failure(e) => None()
+    }
+    function method unwrap_pr(default: T): T
```
robin-aws, line 34: Confused by the naming here - what does "pr" refer to?

examples/Std/OptionAndResult.dfy
```diff
@@ -29,7 +29,7 @@ module Demo {
     m.Put("message", "Hello");
     var o: Option<string> := m.Get("message");
     if o.Some? {
-      print o.get, "\n";
+      print o.unwrap_or, "\n";
```
robin-aws, line 32:
```suggestion
      print o.value, "\n";
```

Otherwise you're printing the `unwrap_or` function.

robin-aws (changes requested):
Just some top-level comments focussing on the style guide, testing approach, and the ever-popular Option and Result. Will do a deeper dive once we're aligned on those.

STYLE.md
```diff
@@ -0,0 +1,336 @@
...
+    && P3()
+````
+
+The same rules apply to `function`, `predicate`, and `lemma` definitions.
+
+## Content Conventions
+
+### Order
+
+Functions, predicates, and methods within a file should be sorted topologically, meaning that everything `method M` depends on should be above `M` in the file.
```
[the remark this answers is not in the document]
parno: I think it's important that the library have a consistent style, which includes ordering.  This isn't meant to impose on how people organize their own Dafny files, but it's hard to imagine a standard library that doesn't have a consistent style.

STYLE.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Naming Convention
+
+Any **classes**, **modules**, **datatypes**, and **newtypes** are named with `PascalCase`.
+
+    Module MyModule {
+        ...
+    }
+    
+Any **lemmas**, **predicates**, **functions**, **methods** and **variables** are named with `snake_case`.
```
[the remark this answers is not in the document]
parno: I guess we interpreted that style guide as internal to Dafny, so we decided to create something fresh for the standard library.  Perhaps we can consider this a strong vote for snake case.

STYLE.md
```diff
@@ -0,0 +1,336 @@
...
+
+    method find_index(arr: seq<int>, k: int)
+        ...
+        
+        
+The **lemma** keyword indicates a ghost method used for proof purposes. Any **lemma** names should be prefixed with `lemma_`.
```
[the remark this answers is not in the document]
parno: We've found it to be a useful convention, since Dafny's various UIs don't make it easy to distinguish ghost and non-ghost code.  In particular, when reading through a complex implementation, it's helpful to be able to see at a glance which calls can potentially mutate state, and which are only being called for proof purposes.

src/Options.dfy
parno: This file was essentially the only file in the repo when we started.  @samuelgruetter created it, so he may be better able to answer questions about it.  Of course, we can take this as an opportunity to clean it up.

lit.site.cfg
parno: Yes, we can probably slim down the `lit.site.cfg` file.  As you say, making sure everything continues to verify seems like the most critical aspect.

src/Options.dfy
samuelgruetter: When I wrote this, I didn't call the module `Options`, but `Std`, with the intention that `Std` would contain the most common "very basic" definitions, and that you could get all of them with one single import.

lit.site.cfg
samuelgruetter: Also, what if a library function has contradictory preconditions and thus is unusable? Or if its postcondition is so weak that nothing useful can be proven with it? IMHO, there should also be basic usability tests that run on CI, in the form of usage examples. That's why I added the following sentence to the README when I started this repo: 

> All added code should be accompanied by some examples in the `examples` directory which illustrates how to use the code.

It might need some work to add examples, but I think it would be well worth doing, because examples that are run on CI are a great form of never-outdated documentation.

STYLE.md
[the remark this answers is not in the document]
robin-aws: I figured that was the motivation. I would love to make ghost state/code easily distinguishable in the UIs. I've noticed that when folks used to more mainstream engineering first start learning Dafny, they often freak out about what looks like extremely expensive implementations, not realizing that the logic in question is erased by the compilers.

STYLE.md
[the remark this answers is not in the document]
robin-aws: Yeah, that style guide has definitely not been aggressively or consistently applied AFAICT. I don't think it was intended to be internal to the Dafny implementation - there's been no effort to standardize all the regression test cases or anything.

@RustanLeino you may want to break the tie here on what you intended. :)

Regardless of the outcome, perhaps we could build a simple refactoring utility to convert identifiers from one style to another, since no matter the outcome we'll have a fair bit of existing Dafny code that will want to migrate styles.

STYLE.md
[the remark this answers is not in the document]
robin-aws: That's fair - I think a lot of these rules we should agree on as universal Dafny style recommendations, but this one could stay as specific to this particular library.

lit.site.cfg
robin-aws: I would drop lit entirely, personally, and just rely on a single call to `dafny` with all source files to compile them to C# (which will ensure they validate), and then run the resulting xUnit tests.

It's probably a good idea to have a separate `tests` directory from `examples` since they serve different purposes. A common pattern I've enforced before is that `tests` is where all the tests go, but each entry in `examples` always gets a simple test that executes it as well.

STYLE.md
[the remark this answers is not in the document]
parno: Yes, a refactoring utility, or some equivalent of `rust-fmt` would be quite nice.

lit.site.cfg
parno: @robin-aws I'm not sure I understand your suggestion.  We still want to make sure the library files verify, which compilation won't help with.  I'm fine with using something other than `lit` for that if you have something concrete in mind.

@samuelgruetter I think one of the benefits of verification is that a method/lemma signature already provides detailed, machine-checkable documentation of the expectations of the inputs/outputs.  If there's something surprising about how a particular method/lemma works, an example could be useful, but I think for most of what we're adding here, the examples would be trivial and uninteresting.  Contributions are welcome, however :)  

> what if a library function has contradictory preconditions and thus is unusable? Or if its postcondition is so weak that nothing useful can be proven with it? 

Yes, that would be worth looking for when reviewing a proposed addition to the library.  Also something users will presumably discover quite quickly if we miss something.  I suspect the kinds of subtle problems that will escape inspection would also fail to be exercised by a trivial example as well.

src/Options.dfy
parno: @samuelgruetter Sorry I was unclear; I meant that comment to apply to all of the questions about this file (see below).  In terms of naming, we could go back to `Std`, but I fear that doesn't communicate much to the reader; i.e., everyone who sees that name will need to open it up to see what's inside. 

@robin-aws We could split `Result` into it's own module, but that starts to seem like we're slicing things pretty thinly.  

Is there another name that would better connote `Option` + `Result`?

lit.site.cfg
robin-aws: Sorry, I could have been clearer with my suggestion. Besides what @samuelgruetter is saying about testing that pre- and post-conditions make sense, I want the CI to ensure the code can be compiled and run and produce the right results. Given that there are lots of `function method`s in this library and not just `function`s, it should be a requirement to ensure that the code is actually compilable and works correctly. 

Here's a really simple example:

```dafny
module SeqTest {
  import Seq
  method {:test} TestApply() {
    var input := [1, 2, 3, 4, 5];
    var output := Seq.apply(i => i * 2, input);
    expect output == [2, 4, 6, 8, 10];
  }
}
```

Building and testing would be a two step process:

```
dafny /compile:0 /spillTargetCode:2 <all Dafny source files>
```

That will still run verification on all the source files and potentially fail. If verification succeeds, you would then execute the resulting C# tests:

```
dotnet test
```

This would require a `.csproj` file similar to the one used in the Dafny test suite: https://github.com/dafny-lang/dafny/blob/master/Test/DafnyTests/DafnyTests.csproj.

The `{:test}` attribute is only handled by the C# compiler for now but we're adding support for the other target languages. Plus, as I mentioned in https://github.com/dafny-lang/dafny/issues/1322#issuecomment-889184653, we should make the process a single simple step by adding something like a `dafny test` command.

src/Options.dfy
sarahc7: Thanks, that was a typo - it's changed to "unwrap_or". I also added a type parameter for the error result.

src/Options.dfy
robin-aws: I want to say `Monads` would make sense, except that I've been explicitly avoiding making that connection since `:-` is explicitly NOT a general-purpose monadic operator, and I'm pessimistic about monadic syntax playing nicely with a language with explicit imperative side-effects (i.e. being able to mutate state with regular statements). We had a lot of discussion about that in https://github.com/dafny-lang/dafny/pull/455

I'm almost happier with `Results` since `Option` is often used to model an optional result, although that's selling `Option` short a bit since it's also useful for optional input.

STYLE.md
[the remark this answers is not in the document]
robin-aws: I confirmed with @RustanLeino offline that he did intend PascalCase as the standard. I also spotted another data point in the built-in datatypes: `iterator` classes automatically define a `MoveNext()` method. :)

robin-aws:
Hey folks! I just wanted to back up and state that I think the only real blocking issue here is the snake_case versus PascalCase issue. The suggestions on better ways to test are something that could be addressed later without breaking any consumers of this library.

@parno Do you have a lot of other code already depending on any of these utilities as written here? If you did have to rename all the symbols, just how painful would it be? We've looked into being able to provide a refactoring utility, but it might be adequate to just do a lot of find-and-replace-ing. I'm pretty firm on the naming convention but I feel bad about blocking things and want to help any way we can. :)

robin-aws:
I should also mention that it would be completely reasonable for the Dafny-to-C++ compiler to start transmuting PascalCase to snake_case, so that at least compiled methods that other C++ code calls can have more idiomatic names.

STYLE.md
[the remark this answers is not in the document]
parno: Since you and Rustan feel that strongly about PascalCase, @sarahc7 has kindly offered to update all of the library code.  One plus is that we can now use `Map` instead of `Apply`, since we will no longer collide with the `map` keyword.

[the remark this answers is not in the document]
robin-aws: Thanks for humouring us, and thank you @sarahc7! :)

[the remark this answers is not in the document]
sarahc7: The library code is now updated to use PascalCase!

parno:
@robin-aws Thanks for the summary above.  @sarahc7 has updated the libraries to use PascalCase, so hopefully we can merge this soon and then consider adding some additional test cases.  We have one project that has already started using these libraries (where we're rewriting to adapt to the PascalCase change), and another that would like to, so it would be nice to try to converge.  @sarahc7 also has some additional libraries for reasoning about sequences of integers and relations between them that we'd like to add once this PR is in.

robin-aws:
Awesome, thanks @parno! I'll have a deeper look at the actual content in the next day or two, but based on skimming it earlier I don't anticipate any other big requests.

src/OptionAndResult.dfy
```diff
@@ -0,0 +1,42 @@
+// RUN: %dafny "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+/*********************************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+**********************************************************************************************/
+
+module OptionAndResult {
```
robin-aws, line 9: For the record I don't think it would be terrible to have separate modules for this, if there's no logical grouping.

FYI I know of a Dafny project that uses the idiom of one definition per module for maximum flexibility, and to avoid repetitive names just calls the datatypes `Option.T` and `Result.T`, which I'm told is an established idiom in OCaml.

src/Collections/Sequences/Seq.dfy
```diff
@@ -0,0 +1,689 @@
...
+  *
+  ***********************************************************/
+  
+  predicate IsPrefix<T>(a: seq<T>, b: seq<T>)
+  {
+    && |a| <= |b|
+    && a == b[..|a|]    
+  } 
+  
+  predicate IsSuffix<T>(a: seq<T>, b: seq<T>) {
```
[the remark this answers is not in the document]
sarahc7: Can `b <= a` be used here? It looks like it returns true if b is a prefix of a, but we want to check if a is a suffix of b.

src/Collections/Sets/Sets.dfy
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /* Construct a set with all integers in the range [a, b). */
+  function method {:opaque} SetRange(a: int, b: int): (s: set<int>)
+    requires a <= b
+    ensures forall i {:trigger i in s} :: a <= i < b <==> i in s
+    ensures |s| == b - a
+    decreases b - a
+  {
+    if a == b then {} else {a} + SetRange(a + 1, b)
```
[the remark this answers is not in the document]
sarahc7: It seems `|s| == b - a` is not obvious if we use `set i: int | a <= i < b`.

src/OptionAndResult.dfy
sarahc7: Would we put `ToResult()` and `ToOption()` in a separate module to avoid a circular dependency?

parno:
@robin-aws Thanks for the detailed review.  I agree that we should make it clear that this library is not in its final state.  The README currently says:
> So, please do use this library, give feedback, and contribute code, but also expect breaking changes.

Do you want something more prominent than that?

Regarding `opaque`, perhaps we can try to provide more explicit guidelines in the STYLE file.  In a perfect world, we wouldn't need `opaque`; Dafny and Z3 would apply exactly the necessary definitions and no more.  Once one starts building non-trivial Dafny developments, this unfortunately proves not to be the case.  In particular, certain styles of functions (e.g., those with recursion or comprehensions) are more likely to cause Dafny/Z3 to explode than others, so marking those as opaque is typically a win for performance and stability.  Nonetheless, I'd be reluctant to give up on automation entirely by marking everything opaque, since that would take us back into a world of manual, interactive theorem proving.

parno:
@robin-aws How are we looking here?  It seems like a lot of suggestions have been incorporated, but it's hard for me to see if there are some that are still pending.  I think we have 1 week of @sarahc7 left, so it would be great to get this in before her REU ends.

robin-aws:
Hi @parno! Apologies for the delay.

I'll have a look at the rest of the code (since I didn't get to the `NonLinearArithmetic` directory yet) and double-check all remaining comments are non-blocking right away.

> > So, please do use this library, give feedback, and contribute code, but also expect breaking changes.
>
> Do you want something more prominent than that?

No I think that's adequate, now that I see that again. I was originally thinking a more explicit pre-release version number like `0.1.0` would be helpful, but since we don't even have a concept of a versioned Dafny package yet I think it can wait.

Regarding `opaque`, your assertion that its use here is at least based on some heuristics is reassuring, so I'm happy with the current state since we're not yet committing to maintaining backwards compatibility.

src/Collections/Sequences/Seq.dfy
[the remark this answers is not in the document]
robin-aws: Ah my mistake, you're correct.

src/Collections/Sequences/Seq.dfy
```diff
@@ -0,0 +1,720 @@
...
+    ensures forall i {:trigger Insert(s, a, pos)[i], s[i]}:: 0 <= i < pos ==> Insert(s, a, pos)[i] == s[i];
+    ensures forall i {:trigger s[i]} :: pos <= i < |s| ==> Insert(s, a, pos)[i+1] == s[i];
+    ensures Insert(s, a, pos)[pos] == a;
+    ensures multiset(Insert(s, a, pos)) == multiset(s) + multiset{a}
+  {
+    assert s == s[..pos] + s[pos..];
+    s[..pos] + [a] + s[pos..]
+  }
+
+  function method {:opaque} Reverse<T>(s: seq<T>): seq<T>
```
robin-aws, line 296: I suppose this is non-blocking, but seems like we should have `ensures forall i | 0 <= i < |s| :: Reverse(s)[i] == s[|s| - i - 1]`

src/Collections/Sequences/Seq.dfy
```diff
@@ -0,0 +1,720 @@
...
+    reveal ToSet();
+    if v !in s then s
+    else
+      var i := IndexOf(s, v);
+      assert s == s[..i] + [v] + s[i+1..];
+      s[..i] + s[i+1..]
+  }
+
+  /* slices out a specific value from the sequence if found */
+  function method {:opaque} RemoveValueOption<T(==)>(s: seq<T>, o: Option<T>): (s': seq<T>)
```
robin-aws, line 264: Hold up, is this function actually useful? I would `if o.Some? then RemoveValue(s, o.Get) else s` would be simpler and fit better into the caller's logic anyway.

Perhaps you were thinking more along the lines of `RemoveValueOption<T(==)>(s: seq<T>, v: T): (s': seq<T>, o: Option<T>)`, along the same lines as `IndexOf` vs. `IndexOfOption` above. Although in that case I'd say just returning a boolean would be better.

src/Collections/Sequences/Seq.dfy
```diff
@@ -0,0 +1,720 @@
...
+      LemmaMultisetHasNoDuplicates(DropLast(s));
+    }
+  }
+
+  /* finds the index of a certain element in the sequence */
+  function method {:opaque} IndexOf<T(==)>(s: seq<T>, v: T): (i: nat)
+    requires v in s
+    ensures i < |s| && s[i] == v
+  {
+    if s[|s|-1] == v then |s| - 1 else IndexOf(s[..|s|-1], v)
```
robin-aws, line 226: Just curious why this indexes from the back with `s[|s|-1]` instead of the front with `s[0]`.

src/Collections/Sequences/Seq.dfy
```diff
@@ -0,0 +1,689 @@
...
+    ensures forall i {:trigger result[i]}:: 0 <= i < |s| ==> result[i] == f(s[i]);
+    reads set i, o {:trigger o in f.reads(s[i])} | 0 <= i < |s| && o in f.reads(s[i]):: o
+  {
+    if |s| == 0 then []
+    else [f(s[0])] + Map(f, s[1..])
+  }
+
+  /* concatenating two sequences and then using "map" is the same as using "map" on each sequence 
+  seperately and then concatenating the two resulting sequences */
+  lemma {:opaque} LemmaMapConcat<T,R>(f: (T ~> R), a: seq<T>, b: seq<T>)
```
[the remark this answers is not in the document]
robin-aws: For the record I realize I should have suggested `LemmaMapDistributesOverConcat` instead, but I can live with this. :)

src/Collections/Sets/Sets.dfy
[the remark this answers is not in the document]
robin-aws: Fair enough!

src/OptionAndResult.dfy
robin-aws: Shoot, good point. Okay, I can live with `OptionAndResult` unless anyone suggests better. :)

src/Collections/Sequences/Seq.dfy
sarahc7: I did this to match `IndexOfOption()`, where indexing from the back gives us the index immediately when we find `v`, so we can directly return `Some(|s| - 1)`.

src/Collections/Sequences/Seq.dfy
sarahc7: That makes sense, I removed this function.

robin-aws (changes requested):
A few more comments - publishing these as I may not get to the rest, but @RustanLeino is intending to review this ASAP as well.

src/Collections/Sequences/Seq.dfy
sarahc7: Ah, makes sense, thanks for pointing this out! I renamed this method `LastIndexOf` and added a separate `IndexOf` method that finds the first occurrence of an element.

jonhnet:
Rustan, I generally agree with your intuition: As someone who's committed to not thinking about triggers, I teach students four methods for responding to timeouts, none of which involve thinking about triggers. However, the standard library seems like it's in the inner circle along with dafny itself; it should be allowed to work necessary magic to better balance automation with timeout protection. What if, not just in the README but in front of every line with a {:trigger}, we add a boilerplate comment that says: ":trigger is an advanced automation control. Avoid its use in user code. Professional drivers on a closed track" to discourage cargo cult programming?

> Dafny does a good job in selecting triggers automatically

It does when it does, which is 99% of the time. But we've learned in big projects that the other 1% is a limiting factor for development scale. This is a bit of a derail, but we need better tooling for diagnosing and correcting timeouts; the current situation is a significant hurdle for Dafny's growth.

RustanLeino:
@parno @Chris-Hawblitzel @jonhnet Thank you for your valuable comments about triggers. It seems we share the intent of using manual triggers only when there's good reason to believe the manual one does better. I like the idea of adding some comment about triggers being an advanced feature (e.g., in the README file). It would then be nice if each manual trigger is accompanied with a comment like those you used in your replies above. (E.g., "Here, Dafny would pick ..., but the intent of the quantifier is ..., so this quantifier is manually given the stricter trigger ...".)

examples/Std/OptionAndResult.dfy
```diff
@@ -1,7 +1,10 @@
-include "../../src/Std.dfy"
+// RUN: %dafny "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+include "../../src/OptionAndResult.dfy"
 
 module Demo {
-  import opened Std
+  import opened OptionAndResult
```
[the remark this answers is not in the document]
RustanLeino: Excellent!

RustanLeino (approved):

robin-aws (approved):

robin-aws:
Thank you so much for this exciting contribution @mmwinchell @sarahc7 and @parno! This is really going to improve the experience when creating a new codebase in Dafny. I'm going to go add it to my favourite Dafny codebase right now. :)

parno:
Thanks very much for all of the thoughtful feedback and help landing this PR!  

And thanks @robin-aws for helping track the remaining issues.

PR #13, opened 2021-09-03
feat: [Wrappers] Adding methods to support :-, covariance on type parameters

robin-aws:


RustanLeino (approved):

parno:
Would it be worth adding a comment (possibly with a link to documentation) about what the covariance `+` means?  It's a feature that even relatively experienced Dafny developers may not have come across before.

robin-aws:
Good call @parno, I'll add another use case to the example file to show what it enables as well.

parno:
Excellent, good idea!

PR #15, opened 2021-09-07
Integer Sequences

sarahc7:
Hi everyone! Thanks for all of your valuable feedback on the `Collections` and `NonlinearArithmetic` sections of the library. We have since added `NatSeq`. Any comments are appreciated :)

Here is the original standard library [discussion](https://github.com/dafny-lang/dafny/issues/1236).


examples/NatSeq/NatSeqExample.dfy
```diff
@@ -0,0 +1,43 @@
...
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+include "../../src/Collections/Sequences/NatSeqConversions.dfy"
+
+module NatSeqExample {
+
+  import opened Uint8_32
```
robin-aws, line 13: I think this is a really cool and clear use case for module refinement in Dafny. Could you also add an example here of refining the general `NatSeqConversions` module with some less common small and large widths, just to highlight that users can do that themselves if they need to?

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+include "Seq.dfy"
+
+abstract module NatSeq {
+
+  import opened DivMod
+  import opened Mul
+  import opened Power
+  import opened Seq
+
+  /* Upper bound of a word */
```
robin-aws, line 27: Throughout, consider "base" or "radix" instead of "bound", and "position" instead of the more overloaded "word"

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+  function method {:opaque} ToNat(xs: seq<uint>): nat
+  {
+    if |xs| == 0 then 0
+    else
+      LemmaMulNonnegativeAuto();
+      ToNat(DropFirst(xs)) * BOUND() + First(xs)
+  }
+
+  /* Converts a sequence to a nat beginning from the most significant word. */
+  function method {:opaque} ToNatRev(xs: seq<uint>): nat
```
robin-aws, line 49: I wouldn't use the "Rev" suffix as I incorrectly assumed it meant this function would behave like `ToNat(Seq.Reverse(xs))`. Perhaps these two functions could be `ToNatLeft` and `ToNatRight`?

In fact, now I'm curious whether these could use `Seq.FoldLeft` and `Seq.FoldRight` instead. If they can't, that feels like an indication we need more lemmas on those Fold functions for them to be actually useful. :)

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+
+/* The first element of a sequence is the least significant word; the last
+element is the most significant word. */
+
+include "../../NonlinearArithmetic/DivMod.dfy"
+include "../../NonlinearArithmetic/Mul.dfy"
+include "../../NonlinearArithmetic/Power.dfy"
+include "Seq.dfy"
+
+abstract module NatSeq {
```
robin-aws, line 20: I admit it took me quite a bit of reviewing before I understood what this module was about. :) I think it could use a comment stating this is specifically about interpreting a sequence of numbers positionally with a given base.

I'd also consider giving it a more suggestive name, perhaps `PositionalNat`?

robin-aws (changes requested):
Oh look, Robin is nit-picking about naming once again. :)

This is really cool stuff, thanks for contributing it! I didn't look too hard at the lemmas just yet as I want to see how my suggestion of using `FoldLeft` and `FoldRight` plays out first.

Out of curiosity, do you also have an implementation of Base64 built on this waiting in the wings too?

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+
+  type uint = i: int | 0 <= i < BOUND()
+
+  //////////////////////////////////////////////////////////////////////////////
+  //
+  // ToNat definition and lemmas
+  //
+  //////////////////////////////////////////////////////////////////////////////
+
+  /* Converts a sequence to a nat beginning from the least significant word. */
```
robin-aws, line 39: I just realized that this is a little-endian interpretation rather than big-endian. I'm not surprised given this code was used in verifying chipsets, but I worry that in a general-purpose library context, most readers will assume sequences are big-endian. That's the natural way to read them in source code: `[1, 5, 3]` with a base of `10` is very naturally interpreted as `153` since that matches how you'd read the literal `153` value.

Would switching this and having to sprinkle lots of `Seq.Reverse` calls in consumers that need little-endianness make it very difficult for those consumers? Could we add an example of using these modules in both endian modes either way?

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+      if xs[i] == ys[i] {
+        reveal ToNatRev();
+        assert DropLast(xs[..i+1]) == xs[..i];
+        assert DropLast(ys[..i+1]) == ys[..i];
+
+        LemmaToNatEqToNatRevAuto();
+        assert ToNat(xs[..i+1]) == ToNatRev(xs[..i+1]);
+      } else {
+        if xs[i] < ys[i]  { LemmaSeqMswInequality(xs[..i+1], ys[..i+1]); }
+        else              { LemmaSeqMswInequality(ys[..i+1], xs[..i+1]); }
```
RustanLeino, line 252: This could be an `else if`

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+        len;
+      }
+    }
+  }
+
+  /* If we start with a nat, convert it to a sequence, and convert it back, we
+  get the same nat we started with. */
+  lemma LemmaNatSeqNat(n: nat)
+    ensures ToNat(FromNat(n)) == n
+    decreases n
```
RustanLeino, line 372: This line is not needed--it is Dafny's default.

src/NativeTypes.dfy
```diff
@@ -0,0 +1,39 @@
...
+
+  const BASE_1:   int := 2
+  const BASE_2:   int := 4
+  const BASE_4:   int := 16
+  const BASE_5:   int := 32
+  const BASE_8:   int := 0x1_00
+  const BASE_16:  int := 0x1_0000
+  const BASE_24:  int := 0x1_000000
+  const BASE_32:  int := 0x1_00000000
+  const BASE_40:  int := 0x1_00000000_00
```
RustanLeino, line 15: Wouldn't it be more natural to write this as `0x100_00000000` (and similar for the others)?

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+    LemmaSeqLen1(DropLast(xs));
+  }
+
+  /* Appending a zero does not change the nat representation of the sequence. */
+  lemma LemmaSeqAppendZero(xs: seq<uint>) 
+    ensures ToNat(xs + [0]) == ToNat(xs)
+  {
+    reveal ToNatRev();
+    LemmaToNatEqToNatRevAuto();
+    calc == {
```
RustanLeino, line 137: `==` is the default default, so you can leave it out from this line.

src/Collections/Sequences/NatSeq.dfy
parno: Good point, that should definitely be documented more explicitly.  In most languages, this would probably be called `BigNat` or something along those lines.  I'm a bit worried that implies a performant implementation, rather than the specification purpose this library is primarily meant for.

src/Collections/Sequences/NatSeq.dfy
parno: Thanks to Intel, we primarily live in a little-endian world, so at least in all of our systems-oriented projects, that's been the dominant mode for reasoning about, say, casts between uint32s and bytes.  We can certainly be more explicit about the fact that the default is little-endian, and we could perhaps expand to offer big-endian variants as well.

robin-aws: This suggests a better name to me, actually: `LittleEndianNat`. That implies it's a positional number anyway, and leaves the door open for a `BigEndianNat`.

parno: Ah, yes, that would make it much clearer.  Good idea.

src/Collections/Sequences/NatSeq.dfy
```diff
@@ -0,0 +1,602 @@
...
+abstract module NatSeq {
+
+  import opened DivMod
+  import opened Mul
+  import opened Power
+  import opened Seq
+
+  /* Upper bound of a word */
+  function method BOUND(): nat
+    ensures BOUND() > 1
```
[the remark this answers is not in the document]
sarahc7: This does seem cleaner, but we'd like to initialize `BOUND` in the refining modules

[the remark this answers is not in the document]
sarahc7: I'm getting an error from this small example:
```
module Test {
  type LessThan10 = x | 0 <= x < 10
  const C: LessThan10
}

module Test' refines Test {
  const C: LessThan10 := 5
}
```
```
Process terminated. Assertion failed.
   at Microsoft.Dafny.Resolver.ResolveTopLevelDecls_Core(List`1 declarations, Graph`1 datatypeDependencies, Graph`1 codatatypeDependencies, Boolean isAnExport) in /private/tmp/dafny-20210205-34509-rqus4y/dafny-3.0.0/Source/Dafny/Resolver.cs:line 2644
   at Microsoft.Dafny.Resolver.ResolveModuleDefinition(ModuleDefinition m, ModuleSignature sig, Boolean isAnExport) in /private/tmp/dafny-20210205-34509-rqus4y/dafny-3.0.0/Source/Dafny/Resolver.cs:line 1002
...
```
Am I missing something?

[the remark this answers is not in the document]
parno: @RustanLeino Did you have thoughts on this issue?  I think this may be the last item holding up this PR.  If function vs. const is a minor concern, perhaps we can merge this PR and then address the bug above and consider converting to a const in a future PR.

parno:
@robin-aws Just to check in, where are we on merging this in?

robin-aws (approved):

robin-aws:
Apologies @parno, I misremembered Rustan's comment as the blocker and forgot I had requested changes.

src/Collections/Sequences/NatSeq.dfy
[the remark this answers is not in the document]
robin-aws: I think that is the right path forward. It's unfortunate since fixing this later will be a breaking change for any refining users, but worth doing while we're still reserving the right to make such changes in these libraries.

I cut https://github.com/dafny-lang/dafny/issues/1481 for the underlying issue and https://github.com/dafny-lang/libraries/issues/21 to track the improvement once that is addressed.

parno:
No problem -- thanks for merging it in!

PR #16, opened 2021-09-24
feat: Need helper function

seebees:
A function to express needed requirements.

There is quite a bit of bikeshedding possible on the names.
And it may be that this should be a function instead of a method.

### DIFF
--- a/src/Wrappers.dfy
+++ b/src/Wrappers.dfy
@@ -87,7 +87,6 @@ module Wrappers {
   // A helper function to ensure a requirement is true at runtime
   // :- Need(5 == |mySet|, "The set MUST have 5 elements.")
   function method Need<E>(condition: bool, error: E): (result: Outcome<E>)
-    ensures result.Pass? ==> condition
   {
     if condition then Pass else Fail(error)
   }
--- a/src/Wrappers.dfy
+++ b/src/Wrappers.dfy
@@ -75,7 +75,7 @@ module Wrappers {
     predicate method IsFailure() {
       Fail?
     }
-    // Note: PropagateFailure returns a Result, not an MUST
+    // Note: PropagateFailure returns a Result, not an Outcome.
     function method PropagateFailure<U>(): Result<U, E>
       requires Fail?
     {
--- a/src/Wrappers.dfy
+++ b/src/Wrappers.dfy
@@ -70,14 +70,14 @@ module Wrappers {
     }
   }
 
-  datatype MUST<E> = NeedsAreMeet | ResultInError(error: E)
+  datatype Outcome<E> = Pass | Fail(error: E)
   {
     predicate method IsFailure() {
-      ResultInError?
+      Fail?
     }
     // Note: PropagateFailure returns a Result, not an MUST
     function method PropagateFailure<U>(): Result<U, E>
-      requires ResultInError?
+      requires Fail?
     {
       Failure(this.error)
     }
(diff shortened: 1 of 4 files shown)


robin-aws (changes requested):
Love it.

And yes, I would make it a function method. Then you can use it in functions like this:

```dafny
function SafeDivide(x: int, y: int): Result<int, string> {
  :- needs(y != 0, "Divide by zero"); 
  Success(x / y)
}
```

Also since you mentioned it a bit offline, don't worry about whether the error value should be an error-producing `() -> e` thunk instead. Since you can only pass expressions as arguments of methods, which are by definition pure and side-effect free*, it will be a safe compiler optimization in the future to only evaluate the second argument if the test fails. This would apply to `expect` statements as well, since they also take an optional `string` argument.

*Disclaimer: in reality this is not 100% true since you could have externally-implemented `function method`s, but that's an orthogonal concern to worry about. :)

src/Wrappers.dfy
```diff
@@ -69,4 +69,26 @@ module Wrappers {
       value
     }
   }
+
+  datatype MUST<E> = NeedsAreMeet | ResultInError(error: E)
```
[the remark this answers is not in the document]
seebees: I was thinking about the dual of option. I think that better reflects what is going on. I see your "But wait, I can get a `Result`" comment, but I would argue that this is the same as `Option`. Because "when we don't need to pass around a reason for the failure," is how an option is described. To me this means that `Result` is the "complete" type and `Option` and `Outcome` are the "lossy" forms. It is just that we tend to prioritize the value so it may feel weird(?)

In the case of `Need` we care about the error, and the transformation into a `Result` is because this "focus on the error" is the same kind of syntactic sugar we are looking at with `Option`. That my pitch anyway :)

src/Wrappers.dfy
```diff
@@ -69,4 +69,26 @@ module Wrappers {
...
+      requires ResultInError?
+    {
+      Failure(this.error)
+    }
+    // Note: no Extract method
+  }
+
+  // A helper function to ensure a requiremnt is me at runtime
+  // :- needs(5 == |mySet|, "The set MUST have 5 elements.")
+  method needs<E>(condition: bool, error: E) returns (result: MUST<E>)
```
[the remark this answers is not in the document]
seebees: I dig it. PascalCase takes me back so. I'll update this as I change it to a `function`

src/Wrappers.dfy
```diff
@@ -69,4 +69,26 @@ module Wrappers {
...
+    function method PropagateFailure<U>(): Result<U, E>
+      requires ResultInError?
+    {
+      Failure(this.error)
+    }
+    // Note: no Extract method
+  }
+
+  // A helper function to ensure a requiremnt is me at runtime
+  // :- needs(5 == |mySet|, "The set MUST have 5 elements.")
```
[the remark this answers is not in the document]
seebees: Added something

src/Wrappers.dfy
```diff
@@ -69,4 +69,26 @@ module Wrappers {
       value
     }
   }
+
+  datatype Outcome<E> = Pass | Fail(error: E)
```
seebees, line 73: @robin-aws What about calling this type `Check`?  A check can pass or fail. And if a check passes then there is no additional information needed. It is only when a check fails that we want to know things.

src/Wrappers.dfy
```diff
@@ -69,4 +69,26 @@ module Wrappers {
...
+    // Note: PropagateFailure returns a Result, not an MUST
+    function method PropagateFailure<U>(): Result<U, E>
+      requires Fail?
+    {
+      Failure(this.error)
+    }
+    // Note: no Extract method
+  }
+
+  // A helper function to ensure a requiremnt is me at runtime
```
keyboardDrummer, line 87: There are a few typo's in the comment.

seebees: thanks fixed

src/Wrappers.dfy
robin-aws: I think that would make less sense, since this is the result of a check rather than the act of checking itself. `Check` would fit better as an alternative to `Need` instead, but I find `Need` fits into the idiom of verbs already in Dafny better so I'm happier with `Need`.

robin-aws (changes requested):
Just some typo fixes, otherwise LGTM!

src/Wrappers.dfy
seebees: Yup, just options.

robin-aws (approved):

robin-aws (approved):

PR #19, opened 2021-10-03
Add LemmaNoDuplicatesCardinalityOfSet

### LINKED ISSUE #18: Need inverse of LemmaCardinalityOfSetNoDuplicates
This lemma proves this implication:
```dafny
  lemma LemmaCardinalityOfSetNoDuplicates<T>(s: seq<T>)
    requires HasNoDuplicates(s)
    ensures |ToSet(s)| == |s|
```

But the opposite implication would also be very useful:
```dafny
  lemma LemmaNoDuplicatesCardinalityOfSet<T>(s: seq<T>)
    requires |ToSet(s)| == |s|
    ensures HasNoDuplicates(s)
```

Especially because `|ToSet(s)| == |s|` is much cheaper to evaluate at runtime than `HasNoDuplicates(s)` would be (if it was a compiled `function method`)

gancherj:
Lemma to fix #18

### DIFF
--- a/src/Collections/Sequences/Seq.dfy
+++ b/src/Collections/Sequences/Seq.dfy
@@ -200,6 +200,28 @@ module Seq {
     }
   }
 
+  /* A sequence with cardinality equal to its set has no duplicates */
+  lemma LemmaNoDuplicatesCardinalityOfSet<T>(s: seq<T>)
+    requires |ToSet(s)| == |s|
+    ensures HasNoDuplicates(s)
+  {
+    reveal HasNoDuplicates();
+    reveal ToSet();
+    if |s| == 0 {
+    } else {
+		  assert s == [s[0]] + s[1..];
+			assert ToSet(s) == {s[0]} + ToSet(s[1..]);
+      if s[0] in s[1..] {
+				assert ToSet(s) == ToSet(s[1..]);
+				LemmaCardinalityOfSet(s[1..]);
+				assert |ToSet(s)| <= |s[1..]|;
+      } else {
+				assert |ToSet(s)| == 1 + |ToSet(s[1..])|;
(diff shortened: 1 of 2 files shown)


src/Collections/Sequences/Seq.dfy
```diff
@@ -200,6 +200,28 @@ module Seq {
...
+    reveal HasNoDuplicates();
+    reveal ToSet();
+    if |s| == 0 {
+    } else {
+      assert s == [First(s)] + DropFirst(s);
+      assert ToSet(s) == {First(s)} + ToSet(DropFirst(s));
+      if First(s) in DropFirst(s) {
+        assert ToSet(s) == ToSet(DropFirst(s));
+        LemmaCardinalityOfSet(DropFirst(s));
+        assert |ToSet(s)| <= |DropFirst(s)|;
```
robin-aws, line 217: Might just be worth a comment that this implies a contradiction with the precondition. AFAICT the contrapositive technique isn't widely used in this repo so it's worth calling it out for the sake of reader understanding.

robin-aws (changes requested):
Thanks @gancherj! I had no idea someone else would jump on this so quickly or I would have assigned it to myself, since I was playing around with it too and came up with an alternative version myself (#20) :) Yours is simpler so I'm happy to accept it instead, although I will update and merge my version since the multiset-related lemmas seem useful independently.

You'll just have to update `Seq.dfy.expect` to expect `70 verified` to get the CI to pass.

src/Collections/Sequences/Seq.dfy
gancherj: Added the comment!

robin-aws (approved):

PR #23, opened 2021-10-06
Add mapValue

hirataqdees:
This commit adds a function `mapValue` for mapping a function over values of a given map

### DIFF
--- a/src/Collections/Maps/Maps.dfy
+++ b/src/Collections/Maps/Maps.dfy
@@ -48,6 +48,17 @@ module Maps {
     m'
   }
 
+  /* Map a function to the values of a map */
+  function method {:opaque} mapValue<K, V, V'>(f: V -> V', m: map<K, V>): (m':map<K, V'>)
+    requires forall k:: k in m.Keys ==> f.requires(m[k])
+    ensures m.Keys == m'.Keys
+    ensures |m| == |m'|
+  {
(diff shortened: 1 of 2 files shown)


src/Collections/Maps/Maps.dfy
```diff
@@ -48,6 +48,17 @@ module Maps {
     m'
   }
 
+  /* Map a function to the values of a map */
+  function method {:opaque} mapValue<K, V, V'>(f: V -> V', m: map<K, V>): (m':map<K, V'>)
```
robin-aws, line 52: Just a few suggestions:

```suggestion
  function method {:opaque} MapValues<K, V, V'>(f: V ~> V', m: map<K, V>): (m': map<K, V'>)
```

* Pascal casing since that's what the style guide says:https://github.com/dafny-lang/libraries/blob/master/STYLE.md#naming-convention :)
* You're already accounting for partial functions in your `requires`, but since you're declaring `f` to have type `V -> V'`, that means `f` is a total function and its precondition is always just "`true`". `V --> V'` is the arrow type that allows for other non-trivial preconditions, and then `V ~> V'` is the even more general arrow type that allows for `reads` clauses as well. You'll have to add a `reads` clause to the function itself of course, something like `reads set o | exists k in m.Keys :: o in f.reads(k)`.

PR #28, opened 2021-11-06
feat: Multiset lemmas

seebees:
A few lemmas with multisets that are useful.

### DIFF
--- a/src/Collections/Multiset/Multiset.dfy
+++ b/src/Collections/Multiset/Multiset.dfy
@@ -1,0 +1,46 @@
+// RUN: %dafny /compile:0 "%s" > "%t"
+// RUN: %diff "%s.expect" "%t"
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+module Multiset {
+
+  /* converts a multiset to a set */
+  function method {:opaque} ToSet<T>(s: multiset<T>): set<T> 
+  {
+    set x: T | x in s
+  }
+
+  /* proves that the cardinality of a multiset is always more than or equal to that
+  of the conversion to a set */
+  lemma LemmaCardinalityOfSetBound<T>(m: multiset<T>)
+    ensures |ToSet(m)| <= |m| 
+  {
+    reveal ToSet();
+    if |m| == 0 {
+    } else {
+      var x :| x in m;
+      var xs := multiset{}[x := m[x]];
+      assert ToSet(xs) == {x};
+      var rest := m - xs;
+      LemmaCardinalityOfSetBound(rest);
+      assert ToSet(m) == ToSet(xs) + ToSet(rest);
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


src/Collections/Sequences/Seq.dfy
```diff
@@ -766,4 +766,43 @@ module Seq {
...
+  }
+
+  /* Every item in the parts to be flattend
+   * is in the flattened seq.
+   */
+  lemma LemmaFlattenMembership<T>(parts: seq<seq<T>>, flat: seq<T>)
+    requires Flatten(parts) == flat
+    ensures forall index
+    | 0 <= index < |parts|
+    :: multiset(parts[index]) <= multiset(flat)
```
robin-aws, line 792:
```suggestion
    ensures forall index | 0 <= index < |parts|
      :: multiset(parts[index]) <= multiset(flat)
```

Easier to read this way I find. And can we likewise indent the other `::` lines below?

src/Collections/Sequences/Seq.dfy.expect
```diff
@@ -1,2 +1,2 @@
 
-Dafny program verifier finished with 71 verified, 0 errors
+Dafny program verifier finished with 73 verified, 0 errors
```
robin-aws, line 2:
```suggestion
Dafny program verifier finished with 74 verified, 0 errors
```

src/Collections/Multiset/Multiset.dfy
```diff
@@ -0,0 +1,46 @@
...
+
+  /* converts a multiset to a set */
+  function method {:opaque} ToSet<T>(s: multiset<T>): set<T> 
+  {
+    set x: T | x in s
+  }
+
+  /* proves that the cardinality of a multiset is always more than or equal to that
+  of the conversion to a set */
+  lemma LemmaCardinalityOfSetBound<T>(m: multiset<T>)
```
robin-aws, line 19:
```suggestion
  lemma LemmaCardinalityOfToSetBound<T>(m: multiset<T>)
```

?

src/Collections/Sequences/Seq.dfy
seebees: I can.  However the reason I left it like that is because of logical chaining.
If you want to say lots of things in either `|` or `::` you end up with lots more white space.

```dafny
predicate Foo(s: seq<int>)
  ensures forall index
  |
    && 0 <= index < |s|
    && magic(s)
    && other(index)
  ::
    && multiset(s[index]) <= multiset(index)
    && things(multiset(s))
    && toProve(index)
{}
```
The reason I like the above, is that it is clear that `|` and `::` fold up the the `ensures` clause.
Much like the `else` lines up with the `if`

```dafny
if condition then
  doWork()
else
  otherWork()
```

robin-aws: That's fair. I'm not convinced I'd want the multi-line formatting for ALL quantifications, even really simple and short ones, but we can argue about that when we get to the auto-formatter :)

RustanLeino:
It seems the only thing stopping an acceptance here is the formatting. For the code that goes into these `libraries`, we should follow the Dafny style guides.

PR #29, opened 2021-11-06
feat: Functional constructs for methods

seebees:
Being able to express logic with map and filter
simplifies the expression of an implementation.

However doing this iteration in a method requires a loop.
Loop invariants are hard.

@robin-aws here is the implementation we have been talking about.
Obviously we need examples before we could merge,
but this can start a discussion about if/how such constructs could be used or simplified.


src/Collections/Sequences/Actions.dfy
```diff
@@ -0,0 +1,261 @@
...
+  import opened Wrappers
+  import opened A = Action
+  import opened Seq
+
+  method Map<A, R>(
+    action: Action<A, R>,
+    s: seq<A>
+  )
+    returns (res: seq<R>)
+    requires forall i | i in s :: action.Requires(i)
```
robin-aws, line 26: (Just food for thought as I THINK we can address this later) - I wonder if we could make this less restrictive, so later elements could depend on earlier ones. Something like:

```
requires |s| > 0 ==> action.Requires(s[0])
requires forall i | 1 <= i < |s| :: 
  (exists r | action.Ensures(s[i - 1], r) ==> action.Requires(s[i]))
```

src/Collections/Sequences/Actions.dfy
```diff
@@ -0,0 +1,261 @@
...
+      invariant |s[..i]| == |rs|
+      invariant forall j ::
+        && 0 <= j < i
+      ==>
+        action.Ensures(s[j], Success(rs[j]))
+    {
+      var maybeR := action.Invoke(s[i]);
+      if maybeR.Failure? {
+        return Failure(maybeR.error);
+      }
```
robin-aws, line 77: Worth a comment about how you're having to work around the issue with `:-` here?

src/Collections/Sequences/Actions.dfy
```diff
@@ -0,0 +1,261 @@
...
+  method ReduceToSuccess<A, B, E>(
+    action: Action<A, Result<B, E>>,
+    s: seq<A>
+  )
+    returns (res: Result<B, seq<E>>)
+    requires forall i | i in s :: action.Requires(i)
+    ensures
+      res.Success?
+    ==>
+      exists a | a in s :: action.Ensures(a, Success(res.value))
```
robin-aws, line 248: Don't you want to make this stronger and say it's the FIRST such element?

src/Collections/Sequences/Actions.dfy
```diff
@@ -0,0 +1,261 @@
...
+    {
+      var r := action.Invoke(s[i]);
+      rs := rs + r;
+      LemmaFlattenConcat(parts, [r]);
+      parts := parts + [r];
+    }
+    return rs, parts;
+  }
+
+  method FlatMapWithResult<A, R, E>(
```
robin-aws, line 124: It's definitely bothering me that we need this and can't just use `Flatten(MapWithResult(action, s))` but I know you warned me it's not that simple. Let me play around with it a bit. :)

robin-aws (changes requested):
I really like the general idea here. Added some initial thoughts and I'm going to play with it a bit myself in the meantime.

src/Collections/Sequences/Actions.dfy
seebees: You should. It may be that with #28 many things are easier.
The difficulty was crossing the flatten boundary.

I will also note, that there may be a runtime efficiency "thing" in here.
If one or the other of these is easier to fold into a lazy seq creation,
then we should go with that one :)

src/Collections/Sequences/Actions.dfy
seebees: Yes, I do! But I did not know how.

MikaelMayer: `exists i | 0 <= i < |s| :: action.Ensures(s[i], Success(res.value)) && forall j | 0 <= j < i ==> exist f | action.Ensures(s[i], Failure(f)) `
 
Maybe there is a better way?

src/Action.dfy
```diff
@@ -0,0 +1,32 @@
...
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+module Action {
+
+  trait {:termination false} Action<A, R>
+  {
+    method Invoke(a: A)
+      returns (r: R)
+      requires Requires(a)
```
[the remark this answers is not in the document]
seebees: Yup. And I think we can have some nice helpers so people can mix in and not have so much typing :)

src/Collections/Sequences/Actions.dfy
cpitclaudel: This looks a bit suspicious.  Should it be parenthesized differently?

```
requires forall i | 1 <= i < |s| :: 
  ((exists r | action.Ensures(s[i - 1], r)) ==> action.Requires(s[i]))
```

Which is then equivalent to 

```
requires forall i | 1 <= i < |s| :: 
  (forall r :: action.Ensures(s[i - 1], r) ==> action.Requires(s[i]))
```

?

src/Action.dfy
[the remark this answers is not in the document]
MikaelMayer: What about having a "modifies" first ?

PR #31, opened 2021-12-02
feat: Add `MapFailure` to `Result`

seebees:
For `:-` to work nicely the Failure side of the `Result` needs to line up with the return value.
While this can be simplified by always using the same type,
this is an unresonalb requirement.

This lets custsomers align the `Failure` side of return value.
AlignFailure may also be a good name.

Example

```dafny
  function Foo(): Result<string, int>

  function Bar(): Result<string, string> {
    var fooString :- Foo().MapFailure(_ => "Foo failed.");
    // work
    Success(fooString)
  }
```

### DIFF
--- a/src/Wrappers.dfy
+++ b/src/Wrappers.dfy
@@ -63,6 +63,13 @@ module Wrappers {
       Failure(this.error)
     }
 
+    function method MapFailure<NewR>(reWrap: R -> NewR): Result<T, NewR>
+    {
+      match this
+        case Success(s) => Success(s)
+        case Failure(e) => Failure(reWrap(e))
+    }
+
     function method Extract(): T
       requires Success?
     {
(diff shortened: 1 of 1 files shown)


src/Wrappers.dfy
```diff
@@ -63,6 +63,13 @@ module Wrappers {
       Failure(this.error)
     }
 
+    function method MapFailure<NewR>(reWrap: R -> NewR): Result<T, NewR>
+    {
+      match this
+        case Success(s) => Success(s)
```
MikaelMayer, line 69: Can you align the `case` with the `match` so that it is consistent with the rest of the file?

MikaelMayer (changes requested):
Once this is fixed we can merge that.

MikaelMayer (approved):

PR #33, opened 2022-02-11
Bump Dafny version to 3.4.0

fabiomadge:
We made a new Dafny release.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,9 +10,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install Dafny
-        uses: tchajed/setup-dafny-action@v1
+        uses: dafny-lang/setup-dafny-action@v1
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -10,9 +10,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install Dafny
-        uses: tchajed/setup-dafny-action@v1
+        uses: dafny-lang/setup-dafny-action@v1
         with:
-          dafny-version: "3.2.0"
+          dafny-version: "3.4.0"
```
robin-aws, line 15: Can we make this a matrix to ensure support of more than just the latest Dafny? Given fixing verification between versions can be tough, it would be great for the standard library to not assume all users can easily move themselves.

fabiomadge: Changing the expect file is a breaking change, so we don't have compatibility to any previous version.

PR #36, opened 2022-02-18
Speed up proof of LemmaToNatLeftEqToNatRight

atomb:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,11 @@
 # This workflow performs static analysis checks.
 name: tests
 
-on: ["pull_request", "push"]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   verification:
@@ -10,9 +14,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install Dafny
-        uses: tchajed/setup-dafny-action@v1
+        uses: dafny-lang/setup-dafny-action@v1
         with:
-          dafny-version: "3.4.1"
+          dafny-version: "3.4.2"
 
       - uses: actions/setup-python@v1
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Install Dafny
         uses: tchajed/setup-dafny-action@v1
         with:
-          dafny-version: "3.2.0"
+          dafny-version: "3.4.1"
 
       - uses: actions/setup-python@v1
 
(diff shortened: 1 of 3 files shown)


src/Collections/Sequences/LittleEndianNat.dfy
```diff
@@ -56,7 +56,7 @@ abstract module LittleEndianNat {
   }
 
   /* Given the same sequence, ToNatRight and ToNatLeft return the same nat. */
-  lemma LemmaToNatLeftEqToNatRight(xs: seq<uint>)
+  lemma {:vcs_split_on_every_assert} LemmaToNatLeftEqToNatRight(xs: seq<uint>)
```
robin-aws, line 59: Since this has no effect on 3.3, we should make this PR bump Dafny up to 3.4 at the same time to be a good atomic change.

robin-aws (changes requested):

src/Collections/Sequences/LittleEndianNat.dfy
atomb: Will do. And I'll also update one other test that's failing with 3.4 since the error message now has slightly more information in it.

src/Collections/Sequences/LittleEndianNat.dfy
```diff
@@ -68,6 +68,7 @@ abstract module LittleEndianNat {
           ToNatLeft(xs);
           Last(xs) * Pow(BASE(), |xs| - 1);
             {
+              reveal Pow();
```
cpitclaudel, line 71: Do you need the lemmas below if you reveal Pow?

atomb: Nope! Good catch.

robin-aws (dismissed):

.github/workflows/tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Install Dafny
         uses: tchajed/setup-dafny-action@v1
         with:
-          dafny-version: "3.2.0"
+          dafny-version: "3.4.1"
```
fabiomadge, line 15: Can we add the other changes to the workflow from #33 as well?

fabiomadge (dismissed):

src/Collections/Sequences/LittleEndianNat.dfy
```diff
@@ -67,10 +67,7 @@ abstract module LittleEndianNat {
         calc {
           ToNatLeft(xs);
           Last(xs) * Pow(BASE(), |xs| - 1);
-            {
-              LemmaPow0Auto();
-              LemmaMulBasicsAuto();
-            }
+            { reveal Pow(); }
```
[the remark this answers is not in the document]
atomb: That works, too, and is perhaps better than revealing `Pow`.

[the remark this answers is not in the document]
atomb: Whoops. It makes it faster but it also makes it fail. :) We could debug it further, but I'm not opposed to leaving the `reveal` in.

robin-aws (approved):

fabiomadge (approved):

PR #38, opened 2022-03-02
Add verification loggers and report generator

robin-aws:
This change adds a best practice we recommend for all Dafny projects: measuring the cost of each individual verification task and setting an upper bound in CI, to guard against future verification instability (see https://github.com/dafny-lang/dafny/issues/1582).

This is done with two steps:

1. Add the `/verificationLogger:csv` parameter when invoking `dafny`, which outputs a CSV file with verification results.
2. Invoke the `dafny-reportgenerator` tool with a configured maximum cost, which will fail the build if any task's cost crosses that threshold.

There are more details on the latter tool here: https://github.com/dafny-lang/dafny-reportgenerator

The good news is that (at least after #36), this codebase is in good shape! :) All tasks take less than 5 seconds across a few runs, so I've set 10 seconds as the upper bound for now. I would prefer to bound the resource count instead since that is a more predictable metric, but Dafny doesn't record the resource count when splitting happens until the upcoming 3.5 release.

Note that because the `/verificationLogger` options cause extra output, I've also changed the lit configuration to no longer assert the exact output of `dafny` when verifying everything. This will make the build more stable by not depending on the exact console output, as all of the code here is always expected to verify successfully.


.github/workflows/tests.yml
```diff
@@ -18,10 +18,20 @@ jobs:
...
+        run: lit --time-tests -v --param 'dafny_params=/verificationLogger:trx /verificationLogger:csv' .
+
+      - name: Generate Report
+        run: find . -name '*.csv' | xargs -t dafny-reportgenerator summarize-csv-results --max-duration-seconds 10
+
+      - uses: actions/upload-artifact@v2  # upload test results
+        if: success() || failure()        # run this step even if previous step failed
+        with:
+          name: verification-results
+          path: 'TestResults/*.trx'
```
atomb, line 37: The build complains that these files aren't found. I think that's because the current logic for report generation chooses the last report type specified, rather than doing all of them. That should probably change.

atomb (commented):
It looks like this won't work with the current version of Dafny since it can't generate both CSV and TRX reports at the same time. That should be easy to fix, though.

.github/workflows/tests.yml
robin-aws: Actually no, Dafny definitely supports both enabling both loggers at once! The issue here is that the path pattern is wrong: initially I was trying to replace `lit` with just one or two top-level calls to `dafny`, in which case you ended up with the results in one top-level `TestResults` folder. But now that I'm leaving `lit` in place, you end up with a `TestResults` folder in every subdirectory instead. :)

I've fixed the pattern and added an option to fail if no files match. Nice catch!

robin-aws: Also notice that unfortunately, the `workflows` trigger won't actually fire until this is merged, which makes it harder to verify. :(

atomb: Ah, got it!

atomb (dismissed):

cpitclaudel:
> Invoke the dafny-reportgenerator tool with a configured maximum cost, which will fail the build if any task's cost crosses that threshold.

Naive question: why is the bounding done by analyzing the CSV after the fact, rather than just passing `timeLimit`? Is that because `/timeLimit` doesn't work at the task level? Or because `/timeLimit` sometimes just doesn't work?

.github/workflows/test-report.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Test Report'
```
cpitclaudel, line 1: `Test` sounds like a verb here to me.  Maybe `Generate test report`?  Same in the file name.

.github/workflows/test-report.yml
```diff
@@ -0,0 +1,16 @@
...
+    types:
+      - completed
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: dorny/test-reporter@v1
+      with:
+        artifact: verification-results
+        name: Verification Tests
```
cpitclaudel, line 14: Why are we using the name "test" here?  Isn't it reporting verification metrics?  Or is it related to Dafny's `:test` somehow?

.github/workflows/tests.yml
```diff
@@ -18,10 +18,21 @@ jobs:
...
       - name: Install lit 
         run: pip install lit OutputCheck
 
-      - name: Verify/Test Dafny code
-        run: lit --time-tests -v .
+      - name: Verify Dafny Code
+        run: lit --time-tests -v --param 'dafny_params=/verificationLogger:trx /verificationLogger:csv' .
+
+      - name: Generate Report
+        run: find . -name '*.csv' | xargs -t dafny-reportgenerator summarize-csv-results --max-duration-seconds 10
```
cpitclaudel, line 31:
```suggestion
        run: find . -name '*.csv' -print0 | xargs -0 --verbose dafny-reportgenerator summarize-csv-results --max-duration-seconds 10
```

cpitclaudel: What happens if there are too many `.csv` reports?

keyboardDrummer:
> > Invoke the dafny-reportgenerator tool with a configured maximum cost, which will fail the build if any task's cost crosses that threshold.
> 
> Naive question: why is the bounding done by analyzing the CSV after the fact, rather than just passing `timeLimit`? Is that because `/timeLimit` doesn't work at the task level? Or because `/timeLimit` sometimes just doesn't work?

I have a related question. If the report-generator is something customers should always use for CI, then isn’t it better to have it be part of the dafny CLI ?

.github/workflows/test-report.yml
robin-aws: Fair point! This is from the template at https://github.com/dorny/test-reporter so I opened a PR to fix it there too. :)

.github/workflows/test-report.yml
robin-aws: Nope, I just did a poor job of editing the template. :) I'm changing it to "Verification Results"

robin-aws:
> > Invoke the dafny-reportgenerator tool with a configured maximum cost, which will fail the build if any task's cost crosses that threshold.
> 
> Naive question: why is the bounding done by analyzing the CSV after the fact, rather than just passing `timeLimit`? Is that because `/timeLimit` doesn't work at the task level? Or because `/timeLimit` sometimes just doesn't work?

I tried to address that in the README for the tool here (https://github.com/dafny-lang/dafny-reportgenerator): "This is better than setting a more aggressive verification cost bound through options like /timeLimit directly, as it allows users to know that their code is still correct, but still blocks code changes that are too expensive to verify and hence likely to break in the future."

.github/workflows/tests.yml
robin-aws: Yeah I'm a bit concerned that will be an issue in the future and cut an issue here: https://github.com/dafny-lang/dafny-reportgenerator/issues/4. The current command length is 1505 so we've got a fair bit of headroom under the 4K limit at least (if it applies to the commands that xargs generates).

robin-aws:
> I have a related question. If the report-generator is something customers should always use for CI, then isn’t it better to have it be part of the dafny CLI ?

Fair question, and especially with the limited functionality the tool is providing so far you could imagine it just being yet more options on the `dafny` CLI. But we expect that it will gain more functionality over time, and the kind of features that can be strongly decoupled from the core functionality of the `dafny` tool. It's also very useful to record the results of verification tasks once and then analyze in multiple ways after the fact, whereas if you had to invoke `dafny` repeatedly you could get different non-deterministic results (at least in terms of duration).

For precedent, the .NET platform also has a similar architecture, where data is written out as a result of options like `--logger` (for test results) and `--collect` (for things like testing coverage), and other tools are used to report on that data in various ways.

keyboardDrummer:
> But we expect that it will gain more functionality over time, and the kind of features that can be strongly decoupled from the core functionality of the `dafny` tool.

What would those features be? When I think of a feature like detecting proofs with a high verification time variability, that seems like something useful to have in Boogie.

> It's also very useful to record the results of verification tasks once and then analyze in multiple ways after the fact, whereas if you had to invoke `dafny` repeatedly you could get different non-deterministic results (at least in terms of duration).

What do you mean by analyse in multiple ways? Do you mean take the measurements from different proof runs and apply different statistics to them? I feel that's something you could do, but I imagine in the general case it would be fine to have Dafny apply a common statistic after which you wouldn't want to do any further analysis.

> For precedent, the .NET platform also has a similar architecture, where data is written out as a result of options like `--logger` (for test results) and `--collect` (for things like testing coverage), and other tools are used to report on that data in various ways.

I'm all for other tools use Dafny's output, but I think that any Dafny related features customers should use in their CI should be in the Dafny CLI.

cpitclaudel (approved):

PR #41, opened 2022-06-03
feat: add Unicode library

alex-chew:
This PR adds basic definitions/functions/lemmas for Unicode values and strings as defined in the Unicode Standard 14.0.0: https://www.unicode.org/versions/Unicode14.0.0/.

(Most/all of this PR is implementation of Chapter 3: https://www.unicode.org/versions/Unicode14.0.0/ch03.pdf)

### DIFF
--- a/src/Unicode/Unicode.dfy
+++ b/src/Unicode/Unicode.dfy
@@ -42,20 +42,19 @@ module Unicode {
     && (p < HIGH_SURROGATE_MIN || p > HIGH_SURROGATE_MAX)
     && (p < LOW_SURROGATE_MIN || p > LOW_SURROGATE_MAX)
 
-  type AssignedCodePoint = p: CodePoint | IsAssignedCodePoint(p) witness *
+  const ASSIGNED_PLANES: seq<bv8> := [
+    0,  // Basic Multilingual Plane
+    1,  // Supplementary Multilingual Plane
+    2,  // Supplementary Ideographic Plane
+    3,  // Tertiary Ideographic Plane
+    14, // Supplementary Special Purpose Plane
+    15, // Supplementary Private Use Area A
+    16  // Supplementary Private Use Area B
+  ]
 
-  predicate {:opaque} IsAssignedCodePoint(i: CodePoint) {
+  predicate {:opaque} IsInAssignedPlane(i: CodePoint) {
     var plane := (i >> 16) as bv8;
-    if plane == 0 then IsInBasicMultilingualPlane(i)
-    else if plane == 1 then IsInSupplementaryMultilingualPlane(i)
-    else if plane == 2 then IsInSupplementaryIdeographicPlane(i)
-    else if plane == 3 then IsInTertiaryIdeographicPlane(i)
-    else if plane == 14 then IsInSupplementarySpecialPurposePlane(i)
-    else if 15 <= plane <= 16 then IsInSupplementaryPrivateUseAreaPlanes(i)
-    else (
-      assert 4 <= plane <= 13;  // planes 4-13 are unassigned
-      false
-    )
+    plane in ASSIGNED_PLANES
   }
 
   // These are actually supersets of the Unicode planes,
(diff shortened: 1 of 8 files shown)


src/Unicode/Utf8EncodingScheme.dfy
```diff
@@ -0,0 +1,67 @@
...
+ * sequence itself.
+ *
+ * Because the UTF-8 encoding form already deals in ordered byte sequences, the UTF-8 encoding scheme is trivial. The
+ * byte ordering is already obvious and completely defined by the UTF-8 code unit sequence itself. The UTF-8
+ * encoding scheme is defined merely for completeness of the Unicode character encoding model.
+ *
+ * (Section 3.10 D95)
+ *
+ * TODO: this should refine an abstract UnicodeEncodingScheme module,
+ * but defining such a module seems impossible so long as dafny-lang/dafny#1639 is unresolved.
```
robin-aws, line 25: Why does https://github.com/dafny-lang/dafny/issues/1639 block this though? Do you have to have a `calc` statement in the abstract module?

src/Unicode/Unicode.dfy
```diff
@@ -0,0 +1,66 @@
...
+  predicate {:opaque} IsInAssignedPlane(i: CodePoint) {
+    var plane := (i >> 16) as bv8;
+    plane in ASSIGNED_PLANES
+  }
+
+  // These are actually supersets of the Unicode planes,
+  // since not all code points in a plane are assigned.
+  //
+  // TODO: correctly check code points against the list of assigned code points
+  // (https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt)
```
robin-aws, line 64: Proving a `forall` would indeed be great, but at a minimum how about a test lemma with sample values from each plane?

src/Unicode/Utf8EncodingScheme.dfy
```diff
@@ -0,0 +1,67 @@
...
+include "Unicode.dfy"
+include "Utf8EncodingForm.dfy"
+
+/**
+ * The Unicode encoding scheme that serializes a UTF-8 code unit sequence in exactly the same order as the code unit
+ * sequence itself.
+ *
+ * Because the UTF-8 encoding form already deals in ordered byte sequences, the UTF-8 encoding scheme is trivial. The
+ * byte ordering is already obvious and completely defined by the UTF-8 code unit sequence itself. The UTF-8
+ * encoding scheme is defined merely for completeness of the Unicode character encoding model.
```
robin-aws, line 20: And also because `unit8` and `bv8` are not the same type? :)

robin-aws (changes requested):
Really love how readable this all is, and how directly it confirms to the terminology in the standard. I made a couple comments about runtime performance, but they aren't blocking - it's extremely useful even just to have a verified reference implementation like this. Really nice work!

src/Unicode/Utf16EncodingForm.dfy
```diff
@@ -0,0 +1,170 @@
...
+    var w := ((firstWord & 0x3C0) >> 6) as bv24;
+    var u := (w + 1) as bv24;
+    var v := (u << 16) | (x1 << 10) | x2 as Unicode.ScalarValue;
+    assert {:split_here} true;
+    assert EncodeScalarValueDoubleWord(v) == m;
+    v
+  }
+
+  //
+  // Test lemmas
```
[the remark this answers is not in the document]
alex-chew: Good call, I've moved all the "tests" into the examples directory.

src/Unicode/Utf8EncodingScheme.dfy
alex-chew: It's not a strict blocker, but I've left a more detailed explanation in the TODO in the latest commits.

src/Unicode/Unicode.dfy
alex-chew: Good call, done in latest commits.

src/Unicode/Utf8EncodingScheme.dfy
alex-chew: Noted in latest commits.

robin-aws (approved):
This is totally $ £ €💰

cpitclaudel (approved):

PR #45, opened 2022-07-19
Fixed duplicate style guides and incorrect indentations

davidcok:
Merged styleguide with the one in the main project; made indentations 2 spaces


STYLE.md
```diff
@@ -1,371 +1,5 @@
...
-  ...
-}
-
-// NO
-lemma LemmaMyLemma <A,B> ( x : seq<seq<A>> , y :B){
-  ...
-}
-```
+(The more or less duplicate copy that was part of the libraries repo has
+been compbined with the project style guide to avoid duplication and divergence.)
```
MikaelMayer, line 5:
```suggestion
been combined with the project style guide to avoid duplication and divergence.)
```

MikaelMayer (changes requested):

MikaelMayer (approved):

robin-aws (commented):
Looks like the only material change between this version and the core one is dropping the "lemmas should start with 'Lemma'" rule.

I'm in support of this, as the IDE is now highlighting all ghost operations/state explicitly, which makes this idiom much less necessary.

@parno any objections here?

RustanLeino (approved):

parno:
Seems like a reasonable change to me, but if that rule is being removed, should all of the existing lemmas in the library have their `Lemma` prefix removed?

PR #46, opened 2022-07-21
feat: Add non-variant version of Wrappers.dfy

robin-aws:

### DIFF
--- a/src/NonVariantWrappers.dfy
+++ b/src/NonVariantWrappers.dfy
@@ -1,0 +1,109 @@
+// RUN: %dafny /compile:0 "%s"
+
+// Assert that this file is minimally different from Wrappers.dfy,
+// to ensure they are mostly drop-in replacements for each other.
+// RUN: %diff "%S/Wrappers.dfy" "%s" | tail -n +3 > "%t" || true
+// RUN: %diff "%t" "%s".diff.expect
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+// Copy of Wrappers.dfy without covariance on type parameters.
+// This version can be useful to work around the lack of variance support
+// for particular compilation targets, especially since variance
+// is not necessary for these datatypes most of the time.
+// E.g. https://github.com/dafny-lang/dafny/issues/2013.
+module Wrappers {
+  
+  datatype Option<T> = None | Some(value: T) {
+    function method ToResult(): Result<T, string> {
+      match this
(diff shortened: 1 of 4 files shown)


src/NonVariantWrappers.dfy
```diff
@@ -0,0 +1,109 @@
...
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+// Copy of Wrappers.dfy without covariance on type parameters.
+// This version can be useful to work around the lack of variance support
+// for particular compilation targets, especially since variance
+// is not necessary for these datatypes most of the time.
```
cpitclaudel, line 16: "most" of the time may be a bit strong (since variance is needed to work with subset types, for example)

src/NonVariantWrappers.dfy
```diff
@@ -0,0 +1,109 @@
+// RUN: %dafny /compile:0 "%s"
+
+// Assert that this file is minimally different from Wrappers.dfy,
+// to ensure they are mostly drop-in replacements for each other.
+// RUN: %diff "%S/Wrappers.dfy" "%s" | tail -n +3 > "%t" || true
+// RUN: %diff "%t" "%s".diff.expect
```
cpitclaudel, line 6: Could we generate this file automatically?  Like we do for Options.txt in the Dafny repo.

src/NonVariantWrappers.dfy
robin-aws: Dialled the statement down a little, but also added an example file to illustrate why covariance often isn't necessary.

src/NonVariantWrappers.dfy
robin-aws: I'm sure we could, but I don't think it's worth it in this case. I don't expect these files to be changed often, and this is a relatively temporary solution until https://github.com/dafny-lang/dafny/issues/2013 is fixed anyway.

cpitclaudel: I'm more convinced by the second argument than the first (I have heaps of stuff to add to `Wrappers.dfy` from `Dafny-in-Dafny`).

But, given the second argument, should we really have this file in the standard library (and not in whichever client code needs it temporarily?)

cpitclaudel (approved):

robin-aws:
I realized this isn't adequate, since other modules in this repo depend on `Wrappers.dfy`, so we need a full clone of the repo instead. Closing this in favour of https://github.com/dafny-lang/libraries-without-variance/pull/1

Issue #47, opened 2022-08-09
Inconsistent use of arrow types


robin-aws:
There is a decent mix of `->`, `-->`, and `~>` arrow types in this repo, owing mostly to personal taste and/or willingness to spend the time generalizing. Ideally all code would be maximally general and use `~>`, but that does lead to more complicated specifications and verification so it is not free.

Should we add this to the style guide? Or even a CI mechanism to insist on `~>`?

MikaelMayer:
Good questions. I think we can always promote `->` to `--> and then to `~>` for arguments when it will be necessary, correct?

cpitclaudel:
Correct @MikaelMayer, but that entails additional proof obligations (the preconditions of the more general function that you're calling).

MikaelMayer:
That's part of the upgrade, but for a user of the library, that wouldn't be a breaking change, would it?

cpitclaudel:
> that wouldn't be a breaking change, would it?

It could be, since the new proofs are harder.
It may be that Dafny managed to call the previous function with the simpler spec and not the one with the more general spec.

For example `Map(f: A -> B, xs: seq<A>)` has no preconditions.  But `Map(f: A --> B, xs: seq<A>)` has `requires forall a | a in xs :: f.requires(a)`.  It should be easy to satisfy this new precondition, but it's still extra work.

Sometimes it's worse, and the most general version of the function has complex preconditions that can be satisfied in various ways that are not immediately obvious (for example, Fold)

Issue #24, opened 2021-10-07
Ensure function methods are efficient at runtime


robin-aws:
Fair warning, this is one of my classic thinking out loud issues that may not be immediately actionable. :)

I'm glad that the majority of the `function` definitions in this repository are `function method`s, so that it is possible for consumers to use them at runtime if they so desire. As per the discussion in https://github.com/dafny-lang/dafny/issues/1492 this is definitely part of the contract of the definition, and it doesn't make sense for the language to support consumers making this decision instead.

However, runtime efficiency is not currently an explicit goal of these libraries, and yet there should be some very low-hanging fruit in terms of easy, classic functional programming optimizations. For example, `Seq.FoldRight` is not tail-recursive, so it could very easily lead to a stack overflow in several target languages: https://github.com/dafny-lang/libraries/blob/162c2a95198f8abd206fcf03cb3c56c672c479b9/src/Collections/Sequences/Seq.dfy#L698-L702

Besides actually improving the implementations here, we need to actually define where the sweet spot should be in terms of runtime efficiency, and add some actual mechanisms in the CI to ensure the expectations are met. This will be tightly coupled to the Dafny compiler and runtime implementations themselves, to ensure the idioms we encourage at least eventually lead to good runtime performance.

For example, repeatedly concatenating small sequences (as most of the functions in `Seq.dfy` do recursively) should be considered an efficient pattern. That is currently only the case when compiling to C# and Java, since they have lazy evaluation of sequence concatenation, but this should be a property of any runtime and therefore not a reason to make the Dafny code more complex to work around it (such as by defining a `List<T> = Nil | Cons(car: T, cdr: List<T>)` datatype to use instead).

PR #49, opened 2022-08-11
adding Merge Sort and Binary Search

prvshah51:


robin-aws (dismissed):
Great start! Here are some top-level suggestions (which will make some of the inline comments I made moot, but I added those anyway just for the sake of learning).

1. This code should be under `src` rather than `examples` - the latter is for sample uses of the actual functions/methods. I would suggest `src/Collections/Arrays/Arrays.dfy` to align with the established layout. The additional total ordering predicates on `compare` would fit nicely into the existing `src/Functions.dfy` file.

2. For sorting, I think proving HeapSort correct will take quite a bit of work. Why not steal the MergeSort implementation I did for the report generator instead? https://github.com/dafny-lang/dafny-reportgenerator/blob/main/src/StandardLibrary.dfy#L38 It definitely won't be as efficient given that it creates lots of intermediate subsequences (especially since most runtimes implement that as a copy instead of a reference), but it's already proven correct, and having that sorting algorithm here will still be an improvement over not having any sorting algorithm. :) You'll just have to replace calls to `f(x) < f(y)` with `compare(x, y)` just as you already have in `BinarySearch`.

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,173 @@
...
+    forall i, j | 0 <= i < j < |a| :: compare(a[i],a[j])
+  }
+
+  lemma LemmaNewFirstElementStillSortedBy<T>(x: T, s: seq<T>,  compare: (T, T) -> bool) 
+    requires SortedBy(s, compare)
+    requires |s| == 0 || compare(x,s[0])
+  {}
+
+function method MergeSortBy<T>(a: seq<T>, compare: (T, T) -> bool ): (result :seq<T>)
+    ensures multiset(a) == multiset(result)
```
atomb, line 128: You also probably need a postcondition that the result is sorted.

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,173 @@
...
+      var leftSorted := MergeSortBy(left, compare);
+      var rightSorted := MergeSortBy(right, compare);
+      
+      MergeSortedBy(leftSorted, rightSorted, compare)
+  }
+
+ function method {:tailrecursion} MergeSortedBy<T>(left: seq<T>, right: seq<T>,  compare: (T, T) -> bool) : (result :seq<T>)
+    requires SortedBy(left,compare)
+    requires SortedBy(right,compare)
+    ensures multiset(left + right) == multiset(result)
```
atomb, line 147: This result should also be sorted.

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,173 @@
...
+
+      [left[0]] +  MergeSortedBy(left[1..], right, compare)
+  }
+
+method Sort<T>(s: seq<T>, compare: (T, T) -> bool) returns (result :seq<T>)
+    ensures multiset(s) == multiset(result)
+    requires StandardLibrary.Transitive(compare)
+    requires Connected(compare)
+ { 
+    MergeSortBy(s,compare)
```
atomb, line 171: Because `Sort` is a method, you'll need something like `result := MergeSortBy(s, compare);`.

src/Collections/Sequences/Sorting.dfy
atomb: In order to add `SortedBy` in a postcondition here, you'll need to export it from the module, as well.

atomb: And I think you'll also need these preconditions:

```dafny
requires StandardLibrary.Transitive(compare)
requires Connected(compare)
```

src/Collections/Sequences/Sorting.dfy
atomb: I think this method will also need the same additional preconditions that `MergeSortBy` does.

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,135 @@
...
+
+  lemma AboutLexicographicByteSeqBelowAux_Connected(x: seq<uint8>, y: seq<uint8>, n: nat)
+    requires n <= |x| && n <= |y|
+    ensures LexicographicByteSeqBelowAux(x, y, n) || LexicographicByteSeqBelowAux(y, x, n)
+    decreases |x| - n
+  {
+  }
+
+method BinarySearch<T>(a: array<T>, key: T,compare: (T, T) -> bool) returns (r : nat)
+  requires forall i,j :: 0 <= i < j < a.Length ==> compare(a[i], a[j])
```
atomb, line 114: You might want to do this to allow the array to contain duplicates.
```suggestion
  requires forall i,j :: 0 <= i < j < a.Length ==> compare(a[i], a[j]) || a[i] == a[j]
```

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,135 @@
...
+  ensures 0 <= r ==> r < a.Length && a[r] == key
+  ensures r < 0 ==> key !in a[..]
+  requires StandardLibrary.Transitive(compare)
+  requires Connected(compare)
```
atomb, line 118: I think this method requires an even more obscure property of the compare function. :)

```dafny
requires StandardLibrary.Trichotomous(compare)
```

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,135 @@
...
+  {
+    var mid := (lo + hi) / 2;
+    if compare(key , a[mid]) {
+      hi := mid;
+    } else if compare(a[mid] , key) {
+      lo := mid + 1;
+    } else {
+      return mid;
+    }
+  }
```
atomb, line 133: At the end here you'll want to return `-1`. Although maybe it makes more sense to have this method return an `Option<nat>` to make things even cleaner, and then you could return `None`.

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,135 @@
...
+
+  lemma AboutLexicographicByteSeqBelowAux_Connected(x: seq<uint8>, y: seq<uint8>, n: nat)
+    requires n <= |x| && n <= |y|
+    ensures LexicographicByteSeqBelowAux(x, y, n) || LexicographicByteSeqBelowAux(y, x, n)
+    decreases |x| - n
+  {
+  }
+
+method BinarySearch<T>(a: array<T>, key: T,compare: (T, T) -> bool) returns (r : nat)
```
atomb, line 113: The way the spec is written right now, this method should return `int` instead of `nat`, though making it return `Option<nat>` might be even nicer, as I mention below.

src/Collections/Arrays/BinarySearch.dfy
atomb: And it turns out that if you have that precondition you don't need `Transitive` or `Connected`.

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,136 @@
+// RUN: %dafny /compile:0 "%s"
+
+include "../../StandardLibrary.dfy"
+include "../../UInt.dfy"
+
+
+
+module BinarySearch{
```
atomb, line 8:
```suggestion
module BinarySearch {
```

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,136 @@
...
+    reveals LexicographicByteSeqBelow, LexicographicByteSeqBelowAux
+    provides AboutLexicographicByteSeqBelow
+    provides StandardLibrary, UInt
+
+  import StandardLibrary
+  import opened UInt = StandardLibrary.UInt
+  import opened Wrappers = StandardLibrary.Wrappers
+  
+
+predicate Reflexive<T(!new)>(R: (T, T) -> bool) {
```
atomb, line 20:
```suggestion
  predicate Reflexive<T(!new)>(R: (T, T) -> bool) {
```

And I'd suggest similar changes throughout this file to ensure the indentation is consistent.

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,181 @@
+// RUN: %dafny /compile:0 "%s"
+
+include "../../StandardLibrary.dfy"
+include "../../UInt.dfy"
+include "../Sequences/Seq.dfy"
+
+module sorting{
```
atomb, line 7:
```suggestion
module Sorting {
```

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,181 @@
...
+    reveals LexicographicByteSeqBelow, LexicographicByteSeqBelowAux
+    provides AboutLexicographicByteSeqBelow
+    provides SortedBy, MergeSortBy, MergeSortedBy
+    provides StandardLibrary, UInt
+
+  import StandardLibrary
+  import opened UInt = StandardLibrary.UInt
+  import opened Seq
+
+predicate Reflexive<T(!new)>(R: (T, T) -> bool) {
```
atomb, line 19: This predicate and a bunch of those after it are duplicated between here and the `BinarySearch` module. Maybe they should go in `StandardLibrary.dfy` until we reorganize it to split it up a little more?

atomb (changes requested):
Looks nice! I just suggested a few small changes for consistency and removal of duplication.

Eventually, I suspect we should split up `StandardLibrary.dfy` to put some of its contents in the more specialized modules that already exist (and maybe some new ones). For now, though, this at least makes a bunch more definitions available.

src/UInt.dfy
```diff
@@ -0,0 +1,336 @@
+// RUN: %dafny /compile:0 "%s"
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+module StandardLibrary.UInt {
+
+  // TODO: Depend on types defined in dafny-lang/libraries instead
```
atomb, line 10: This comment probably isn't needed anymore. :)

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,41 @@
...
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+include "../../StandardLibrary.dfy"
+include "../../UInt.dfy"
+
+module BinarySearch {
+  export
+    provides StandardLibrary, UInt
```
atomb, line 13: You'll probably want `provides BinarySearch`, too, so that people can use it.

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,41 @@
...
+
+  import StandardLibrary
+  import opened UInt = StandardLibrary.UInt
+  import opened Wrappers = StandardLibrary.Wrappers
+
+method BinarySearch<T>(a: array<T>, key: T, compare: (T, T) -> bool) returns (r:Option<nat>)
```
atomb, line 19: It'd be nice to have this indented two spaces to be consistent with the imports above it.

atomb (dismissed):
Nicely done!

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,41 @@
...
+include "../../StandardLibrary.dfy"
+
+module Search {
+  export
+    provides BinarySearch, Wrappers, StandardLibrary
+  
+  import StandardLibrary
+  import opened Wrappers = StandardLibrary.Wrappers
+
+  method BinarySearch<T>(a: array<T>, key: T, compare: (T, T) -> bool) returns (r:Option<nat>)
```
MikaelMayer, line 17: Compare function usually return an int: negative for <, zero for == and positive for >. Or a datatype
If you a returning a bool, you should name this function "lessThan", "lessEqThan", "greaterThan" or "greaterEqThan", depending on what this means.

But my preference is really that you return a datatype [like this](https://github.com/dafny-lang/compiler-bootstrap/commit/d9a5734a47b9dd78b3af764ea913e7a19bc8c857#diff-2abd45eb58cd25c1bff3a753036307c5bc25d3309d470e9e45ea73d0308323acR588) and use the result in an

```
datatype ComparisonResult = LessThan | Equals | GreaterThan

var comp :=  compare(a[mid] , key);
match comp {
case LessThan =>
case Equals =>
case GreaterThan =>
}
```

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,191 @@
...
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+include "../../StandardLibrary.dfy"
+include "../../UInt.dfy"
+include "../Sequences/Seq.dfy"
+
+module Sorting {
```
MikaelMayer, line 12: Could you use the attribute `{:options "/functionSyntax:4"}` and change all your `predicate method` to `predicate` ?

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,191 @@
...
+    reveals LexicographicByteSeqBelow, LexicographicByteSeqBelowAux
+    provides AboutLexicographicByteSeqBelow
+    provides SortedBy, MergeSortBy, MergeSortedBy
+    provides StandardLibrary, UInt
+
+  import StandardLibrary
+  import opened UInt = StandardLibrary.UInt
+  import opened Seq
+
+  predicate Reflexive<T(!new)>(R: (T, T) -> bool)
```
MikaelMayer, line 24: Could you put Reflexive, AntiSymmetric, Transitive, etc. in a file named `Relations.dfy` instead of "standardLibrary"? Here this is too generic for "Sorting", and this entire project is supposed to be the standard library, so there is no point having a file named "standard library", its name should more reflect what it defines.

MikaelMayer (changes requested):
It will be good to have it ! Almost there.

cpitclaudel:
This is great.  Following up on what Mikael said:

- It would be nice to separate things out more cleanly: relations, sorting, and searching
- You can take the generic comparison type from https://github.com/dafny-lang/compiler-bootstrap/commit/d9a5734a47b9dd78b3af764ea913e7a19bc8c857 (`datatype Cmp`)
- Down the line it would be nice to support sorting reference types, but we can address that later
- I worry about the performance (sorting sequences instead of arrays).  In particular, I think the current code has complexity (at least in C#) `n² log(n)`, but we can probably add a "by method" later
- The main entry point probably doesn't need to be a method

It would be nice to add docstrings on the top functions, too

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,191 @@
...
+      [right[0]] + MergeSortedBy(left, right[1..], compare)
+
+    else
+      LemmaNewFirstElementStillSortedBy(left[0], MergeSortedBy(left[1..], right, compare), compare);
+      assert left == [left[0]] + left[1..];
+
+      [left[0]] +  MergeSortedBy(left[1..], right, compare)
+  }
+
+  method Sort<T>(s: seq<T>, compare: (T, T) -> bool) returns (result :seq<T>)
```
MikaelMayer, line 184:
```suggestion
  function method {:opaque} Sort<T>(s: seq<T>, compare: (T, T) -> bool) returns (result :seq<T>)
```

src/Collections/Sequences/Sorting.dfy
```diff
@@ -0,0 +1,191 @@
...
+  {
+    forall i, j | 0 <= i < j < |a| :: compare(a[i],a[j])
+  }
+
+  lemma LemmaNewFirstElementStillSortedBy<T>(x: T, s: seq<T>,  compare: (T, T) -> bool) 
+    requires SortedBy(s, compare)
+    requires |s| == 0 || compare(x,s[0])
+  {}
+
+  function method MergeSortBy<T>(a: seq<T>, compare: (T, T) -> bool ): (result :seq<T>)
```
MikaelMayer, line 138:
```suggestion
  function method MergeSortBy<T>(a: seq<T>, compare: (T, T) -> bool ): (result :seq<T>)
    // Split a sequence in two, sorts the two subsequences using itself, and merge the two sorted sequences using `MergeSortedBy`
```

src/Lexicographics.dfy
```diff
@@ -0,0 +1,95 @@
...
+    export 
+    reveals LexicographicByteSeqBelow, LexicographicByteSeqBelowAux 
+    provides Helpers, UInt, Relations
+
+    import Helpers = Relations.Helpers
+    import opened UInt = Helpers.UInt
+    import opened Relations
+
+  // reflexivelexicographical comparison of byte sequences
+  predicate method LexicographicByteSeqBelow(x: seq<uint8>, y: seq<uint8>)
```
MikaelMayer, line 20: Remove "Lexicographic" from all the predicate, lemmas and method names unless strictly necessary, since you are in the Lexicographics module.

src/Lexicographics.dfy
```diff
@@ -0,0 +1,95 @@
+// RUN: %dafny /compile:0 "%s"
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+include "Relations.dfy"
+
+module Lexicographics {
```
MikaelMayer, line 10: I would rename this module to "LexicographicOrder"

src/Helpers.dfy
```diff
@@ -0,0 +1,300 @@
...
+    ==  { reveal ASymmetric; }
+      k0 == |a| && k1 == |b|;
+    ==>  { assert |a| == k0 <= max && |b| == k1 <= max ==> k0 == k1; }
+      max == |a| == |b|;
+    ==>  { reveal SameUntilMax; }
+      a == b;
+    }
+  }
+
+  lemma LexIsTransitive<T>(a: seq<T>, b: seq<T>, c: seq<T>, less: (T, T) -> bool)
```
MikaelMayer, line 154: Can you move this and everything Lexicographic related to the lexicographics module?
You could always define another module, like Lexicographics.ProofHelpers

Also, please rename "less" to "lessThan".
Consider (but think carefully if it will be better) to change "less" to "compare" that returns a ComparisonResult, for consistentcy

src/Helpers.dfy
```diff
@@ -0,0 +1,300 @@
...
+*******************************************************************************/
+
+include "../../libraries/src/Wrappers.dfy"
+include "UInt.dfy"
+
+module Helpers {
+  import opened Wrappers
+  import opened U = UInt
+
+  function method Min(a: int, b: int): int {
```
MikaelMayer, line 15: This ought to be in  Math module in another ifle.

src/Helpers.dfy
```diff
@@ -0,0 +1,300 @@
...
+    if a < b then a else b
+  }
+
+  function method Fill<T>(value: T, n: nat): seq<T>
```
MikaelMayer, line 19: This should be in the file Seq.dfy

src/Helpers.dfy
```diff
@@ -0,0 +1,300 @@
...
+    ensures |Fill(value, n)| == n
+    ensures forall i :: 0 <= i < n ==> Fill(value, n)[i] == value
+  {
+    seq(n, _ => value)
+  }
+
+  method SeqToArray<T>(s: seq) returns (a: array)
```
MikaelMayer, line 26: In the file Seq.dfy

src/Helpers.dfy
```diff
@@ -0,0 +1,300 @@
...
+    // "Fresh" expressions require editing memory
+    ensures fresh(a)
+    ensures a.Length == |s|
+    ensures forall i :: 0 <= i < |s| ==> a[i] == s[i]
+  {
+    a := new T[|s|](i requires 0 <= i < |s| => s[i]);
+  }
+
+  lemma SeqPartsMakeWhole<T>(s: seq<T>, i: nat)
```
MikaelMayer, line 35: In the file Seq.dfy

src/Collections/Sequences/Sorting.dfy
MikaelMayer: This again

src/Collections/Sequences/Sorting.dfy
MikaelMayer: This again

src/Collections/Sequences/Sorting.dfy
MikaelMayer: This again

src/Collections/Arrays/BinarySearch.dfy
MikaelMayer: This again

MikaelMayer (changes requested):
This PR is in a better shape!

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -25,17 +30,15 @@ module Search {
...
-        return Some(mid);
-      }
-    }
+    var mid := (lo + hi) / 2;
+
+    var comp := compare(key , a[mid]);
+    match comp
+      case LessThan => hi := mid;
+      case GreaterThan => lo:= mid + 1;
+      case _ => return Some(mid);
```
MikaelMayer, line 39: Very elegant! Can you please indent this code?

src/Collections/Sequences/Seq.dfy
```diff
@@ -873,19 +882,19 @@ module Seq {
   ***********************************************************/
 
 
-  predicate method SortedBy<T>(a: seq<T>)
+  predicate method SortedBy<T>(a: seq<T>,compare: (T, T) -> Comparison.CompResult)
```
MikaelMayer, line 885:
```suggestion
  predicate method SortedWith<T>(a: seq<T>, compare: (T, T) -> Comparison.CompResult)
```

src/CompariosRelations.dfy
```diff
@@ -0,0 +1,44 @@
+include "Helpers.dfy"
```
MikaelMayer, line 1: This file is named "CompariosRelations.dfy" ^^

src/CompariosRelations.dfy
```diff
@@ -0,0 +1,44 @@
...
+module Relations.ComparisonsRelation { 
+
+    export 
+    reveals Reflexive, AntiSymmetric, Connected, TotalOrdering
+    import opened Helpers.Comparison
+
+
+    predicate Reflexive<T(!new)>(R: (T, T) -> CompResult) 
+    {
+        forall x :: R(x, x)
```
MikaelMayer, line 12: How can this typecheck if R returns a CompResult?

src/Helpers.dfy
```diff
@@ -5,296 +5,12 @@
...
-            }
-          }
-        }
-      }
-    }
-  }
 }
 
-module Comparison {
+module Helpers.Comparison {
```
MikaelMayer, line 14: Can you please move this to Relations or somewhere else? That way you can delete this  now insignificant and mostly empty file

src/LexicographicHelpers.dfy
```diff
@@ -0,0 +1,243 @@
...
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+include "Relations.dfy"
+include "Lexicographics.dfy"
+include "UInt.dfy"
+
+module LexicographicHelpers{
```
MikaelMayer, line 12: Could you please rename this `LexicographicSequence`, so that you can remove all the "Lex" prefix from all the procedures below?

src/Lexicographics.dfy
MikaelMayer: I would still rename this to "LexicographicByteSequence".

src/Lexicographics.dfy
```diff
@@ -6,90 +6,109 @@
...
+   * ordering `less` on the elements of these sequences, determine whether or not
+   * `a` is lexicographically "less than or equal to" `b`.
+   *
+   * `a` is lexicographically "less than or equal to" `b` holds iff
+   *   there exists a `k` such that
+   *   - the first `k` elements of `a` and `b` are the same
+   *   - either:
+   *      -- `a` has length `k` (that is, `a` is a prefix of `b`)
+   *      -- `a[k]` is strictly less (using `less`) than `b[k]`
+   */
```
MikaelMayer, line 110: This comment should be moved to the right module.

src/Relations.dfy
```diff
@@ -1,12 +1,6 @@
...
-
 module Relations { 
     export 
-    reveals Reflexive, AntiSymmetric, Connected, TotalOrdering
-    provides Helpers
-
-    import opened Helpers
+    reveals Reflexive, AntiSymmetric, Connected, TotalOrdering, Transitive, Trichotomous
 
     predicate Reflexive<T(!new)>(R: (T, T) -> bool)
```
MikaelMayer, line 5: I prefer that we unify our predicates and name them `IsReflexive`, `IsTransitive`, etc.

src/Relations.dfy
```diff
@@ -27,8 +21,29 @@ module Relations {
     {
```
MikaelMayer, line 21: IsTotalOrder

src/LexicographicHelpers.dfy
```diff
@@ -0,0 +1,243 @@
...
+    export 
+        provides LexIsReflexive, LexIsAntisymmetric, LexIsTransitive, LexIsTotal, UInt8LessIsTrichotomousTransitive
+        provides ThereIsAMinimum, MinimumIsUnique, FindMinimum
+        provides SetToOrderedSequence, IsMinimum
+        provides UInt, Relations, Lexicographics
+        provides  LessOrEqual, LessOrEqualAux
+
+    import opened Relations
+    import opened Lexicographics
```
MikaelMayer, line 21: Why do you need Lexicographics here? I thought Lexicographic sequence bytes would to the contrary inherit from Lexicographic sequences that you are describing there.

src/LexicographicHelpers.dfy
```diff
@@ -0,0 +1,243 @@
...
+    import U = UInt
+    import opened Lexicographics.UInt
+
+    predicate method LessOrEqual<T(==)>(a: seq<T>, b: seq<T>, less: (T, T) -> bool) {
```
MikaelMayer, line 25: Same here. Could you please rename this predicate to `IsLessOrEqual`?

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -6,17 +6,22 @@
...
-  method BinarySearch<T>(a: array<T>, key: T, compare: (T, T) -> bool) returns (r:Option<nat>)
-    requires forall i,j :: 0 <= i < j < a.Length ==> compare(a[i], a[j]) || a[i] == a[j]
-    requires Helpers.Trichotomous(compare)
+    import Helpers
+    import opened Relations
+    import opened Wrappers
+    import opened Comparison = Helpers.Comparison
+
+  method BinarySearch<T>(a: array<T>, key: T, compare: (T, T) -> Comparison.CompResult) returns (r:Option<nat>)
+    requires forall i,j :: 0 <= i < j < a.Length ==> compare(a[i], a[j]).LessThan? || a[i] == a[j]
```
MikaelMayer, line 23:
```suggestion
    requires forall i,j :: 0 <= i < j < a.Length ==> a[i] == a[j] || compare(a[i], a[j]).LessThan?
```
That reduces the number of comparisons if elements are mostly equal.

MikaelMayer (changes requested):
I like this shape better ! A few more comments there and there to make it suitable for a library.

src/Collections/Sequences/Seq.dfy
keyboardDrummer: Why the preference for `With` instead of `By` ?

src/LexicographicHelpers.dfy
prvshah51: it was creating cyclic dependencies with imports, so I need to sort it out.

src/Lexicographics.dfy
prvshah51: I will rename "Lexicographics" to a better describing name ,I just wanted to make sure there is no other file changes before doing that.

src/UInt.dfy
```diff
@@ -0,0 +1,335 @@
+// RUN: %dafny /compile:0 "%s"
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+module Helpers.UInt {
+
+  newtype uint8 = x | 0 <= x < 0x100
```
MikaelMayer, line 10: Would it be possible to move all these definitions to BoundedInts.dfy ? I did not see that this file existed, and that would be handy as well if you could incorporate the definitions of this upcoming PR:

https://github.com/dafny-lang/libraries/pull/51/files#diff-437e3168477b72c1dc1b4f5e935632c1f00659c1497c7f8d673ec2de1a1efa69R36

src/CompariosRelations.dfy
atomb: This raises the interesting question that these relations are awkward to define (if possible at all) for relations that return `CompResult` (and therefore aren't really relations in the normal mathematical sense). I tried rewriting `Trichotomous` for `CompResult` to see what it'd look like (since I think it may be a sufficient precondition for both searching and sorting) and came up with this, which feels like a bit of a mess in relation to the original:

```
predicate Trichotomous<T(!new)>(compare: (T, T) -> CompResult) 
{
    (forall x, y :: compare(x, y).LessThan? || compare(x, y).Equals? || compare(y, x).LessThan?) &&  // at least one of the three using less
    (forall x, y :: compare(x, y).GreaterThan? || compare(x, y).Equals? || compare(y, x).GreaterThan?) &&  // at least one of the three using greater
    (forall x, y :: compare(x, y).LessThan? && compare(y, x).LessThan? ==> false) &&  // not both less
    (forall x, y :: compare(x, y).GreaterThan? && compare(y, x).GreaterThan? ==> false) &&  // not both greater
    (forall x, y :: compare(x, y).Equals? ==> x == y) && // Equals means equality
    (forall x, y :: compare(x, y).LessThan? ==> x != y) && // not less and equal
    (forall x, y :: compare(x, y).GreaterThan? ==> x != y)  // not greater and equal
}
```

Maybe using a less-than relation is actually a better approach in Dafny, given how well-behaved it is mathematically?

atomb: Alternatively, we could adopt all of the code [from here](https://github.com/dafny-lang/compiler-bootstrap/blob/cpitclaudel_entities-impl/src/Utils/Library.dfy#L642-L748) directly.

atomb: Yes, having considered it more, let's adopt that `compiler-bootstrap` code unless it leads to other issues.

cpitclaudel: Yep, defining `.Le` and taking about it is the way to go, I think.
But note that in general for sorting you don't need to know that `Equals` implies `==` (and, in fact, you don't want it, either, since you may sort according to a projection of the value like `function Le(a: (int, int), b: (int, int)) { a[0] <= b[0] }`)

atomb: Yep, both good points (and further arguments for using your code!).

src/Collections/Arrays/BinarySearch.dfy
robin-aws: Nah I wouldn't prematurely optimize like that - it also increases the number of comparisons if the elements are mostly not equal (which I would expect to be more common).

In fact this is ghost code, so why do we even care about extra operations? :)

src/Relations/Comparison.dfy
```diff
@@ -0,0 +1,127 @@
...
+
+    lemma TotalValid(ts: Seq<T>)
+      ensures Total?(ts) ==> Valid?(ts)
+    {
+      reveal Total?();
+      reveal Valid?();
+    }
+  }
+        /*
+   * Here is an example relation and a lemma that says the relation is appropriate for use in
```
robin-aws, line 118: Let's move this to `examples` instead then.

src/LexicographicHelpers.dfy
```diff
@@ -0,0 +1,246 @@
+// RUN: %dafny /compile:0 "%s"
```
robin-aws, line 1: I'd like to leave this file out of this PR for now too. It's more relevant since it's providing a useful instance of ordering, but we can add that later.

src/Lexicographics.dfy
```diff
@@ -0,0 +1,96 @@
+// RUN: %dafny /compile:0 "%s"
```
robin-aws, line 1: Similarly, omit this file from this PR for now.

src/BoundedInts.dfy
```diff
@@ -34,5 +34,348 @@ module BoundedInts {
   newtype nat16 = x: int  | 0 <= x < 0x8000
   newtype nat32 = x: int  | 0 <= x < 0x8000_0000
   newtype nat64 = x: int  | 0 <= x < 0x8000_0000_0000_0000
+
```
robin-aws, line 37: I think we can omit all these changes once we remove the other files I'm suggesting we remove. We may just want to define one or two comparison functions in `examples` files instead.

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,46 @@
...
+
+    import opened Relations
+    import opened Wrappers
+    import opened Comparison = Relations.Comparison
+
+  method BinarySearch<T>(a: array<T>, key: T, compare: (T, T) -> Comparison.CompResult) returns (r:Option<nat>)
+    requires forall i,j :: 0 <= i < j < a.Length ==> compare(a[i], a[j]).LessThan? || a[i] == a[j]
+    requires Total?(compare)
+    ensures r.Some? ==> r.value < a.Length && a[r.value] == key
+    ensures r.None? ==> key !in a[..]
```
robin-aws, line 25: Some versions of binary search will return the index where the element WOULD be when it isn't present. If we want to support that here in the future we'd have to change the return type which would be a breaking change, or define a second version. We could use `Result<nat, nat>`, but that definitely wouldn't read as naturally as a more generic `Either<A, B>` type.

src/Collections/Sequences/MergeSort.dfy
```diff
@@ -0,0 +1,70 @@
...
+  *
+  *  Sorting of Sequences
+  *
+  ***********************************************************/
+
+module Seq.MergeSorting {
+
+    import opened Comparison = Relations.Comparison
+ 
+  predicate method SortedBy<T>(a: seq<T>,compare: (T, T) -> CompResult)
```
robin-aws, line 13: This and the following lemma should be moved to `Comparison.dfy` since it's not specific to any particular sorting algorithm.

src/Collections/Sequences/MergeSort.dfy
```diff
@@ -0,0 +1,70 @@
...
+  {
+    forall i, j | 0 <= i < j < |a| :: compare(a[i],a[j]).LessThan?
+  }
+
+  lemma LemmaNewFirstElementStillSortedBy<T>(x: T, s: seq<T>,  compare: (T, T) -> CompResult) 
+    requires SortedBy(s, compare)
+    requires |s| == 0 || compare(x,s[0]).LessThan?
+  {}
+
+  //Splits a sequence in two, sorts the two subsequences using itself, and merge the two sorted sequences using `MergeSortedBy`
```
robin-aws, line 23:
```suggestion
  //Splits a sequence in two, sorts the two subsequences using itself, and merge the two sorted sequences using `MergeSortedWith`
```

robin-aws (changes requested):
This is looking really nice! Love to see the team effort here. :)

I made a bunch of top level suggestions intended to trim the scope of the PR way down. There's a ton of potentially useful code in here from various sources, but I'd like to first polish and merge just the generic merge sort and binary search functions and the bare minimum of their dependencies first. We can add more useful instances of orderings in follow-up PRs.

src/Collections/Sequences/Seq.dfy
```diff
@@ -13,13 +13,22 @@
...
-include "../../Wrappers.dfy"
 include "../../Math.dfy"
+include "../../Wrappers.dfy"
+include "../../UInt.dfy"
+include "../../Relations/Relations.dfy"
 
 module Seq {
+  export 
+   // provides SortedBy, MergeSortBy, MergeSortedWith
+    provides Relations//, Comparison
```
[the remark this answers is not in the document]
prvshah51: oh, I didn't know, will you suggest to just remove export set from all the files I added to ?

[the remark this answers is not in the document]
robin-aws: If possible, yup. I think it will be a moot point once you remove all the edits I suggested though.

src/Collections/Sequences/Seq.dfy
```diff
@@ -765,4 +783,160 @@ module Seq {
...
+
+  //     [left[0]] +  MergeSortedWith(left[1..], right, compare)
+      
+
+  //     else
+  //       LemmaNewFirstElementStillSortedBy(right[0], MergeSortedWith(left, right[1..], compare), compare);
+  //       assert right == [right[0]] + right[1..];
+
+  //     [right[0]] + MergeSortedWith(left, right[1..], compare)
+  // }
```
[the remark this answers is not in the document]
robin-aws: I don't feel too strongly either way, but I think it would be fine to just add it to `Seq.dfy` since the first two declarations are moving to `Comparison.dfy`, so it's just two more function methods.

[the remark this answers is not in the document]
prvshah51: Just to update, I had to move `Seq.MergeSort` out of Seq.dfy and make a new file as it needs `Comparison` included and on the other hand `Comparison` needs `Seq` included and that was creating duplicate module included `Seq` resolver error in the file Seq.dfy

src/Collections/Arrays/BinarySearch.dfy
MikaelMayer: > In fact this is ghost code, so why do we even care about extra operations? :)

Right, that's a good point :-)

src/Collections/Arrays/BinarySearch.dfy
robin-aws: I've decided I'm fine with having to define a second alternate version if and when we want this functionality.

robin-aws:
@prvshah51 I've pushed several commits to remove more code we don't need (yet!), and made a few other changes to get it ready to merge IMO:

* After discussion with @atomb I've removed `Comparison.dfy` and stuck to plain old `(T, T) -> bool` relations, which simplifies things for now.
* Moved `Relations.dfy` to the top-level `src` directory (since we also have `Functions.dfy` there already)
* Added an example that sorts and then searches, which also revealed that the "sorted" precondition on BinarySearch wasn't correct!

Let me know if you have any concerns, and @atomb / @MikaelMayer I'll leave it to you two to approve since I feel a little biased now that I've thrown my own commits on this. :)

robin-aws:
@MikaelMayer I just noticed you had some style requests that were lost in the shuffle. I'll respond to the ones I see that still seem relevant, but please just provide a quick comment at the top level if there are others we missed.

Re: the naming convention for predicates, e.g. "IsTotalOrder" instead of "TotalOrder", that's potentially a good idea but the latter convention is already established here in predicates like `Functions.Injective`. I'd suggest if you still feel strongly about it that you cut a repo issue for discussion, similar to #47.

Re: "lessThan" instead of "less", I don't have a strong opinion so I'd be willing to change it, personally. @prvshah51?

prvshah51:
> Re: "lessThan" instead of "less", I don't have a strong opinion so I'd be willing to change it, personally. @prvshah51?
Both options are fine with me.

src/Collections/Arrays/BinarySearch.dfy
MikaelMayer: Yes, like `BinarySearchInsertionIndex`, would be good down the road.

src/Collections/Sequences/MergeSort.dfy
```diff
@@ -0,0 +1,57 @@
...
+*******************************************************************************/
+
+include "../../Relations.dfy"
+
+module Seq.MergeSort {
+
+  import opened Relations
+
+  //Splits a sequence in two, sorts the two subsequences using itself, and merge the two sorted sequences using `MergeSortedWith`
+  function method MergeSortBy<T>(a: seq<T>, compare: (T, T) -> bool): (result :seq<T>)
```
MikaelMayer, line 15:
```suggestion
  function method MergeSortBy<T>(a: seq<T>, lessEqThan: (T, T) -> bool): (result :seq<T>)
```

src/Relations.dfy
```diff
@@ -0,0 +1,72 @@
...
+  /*
+  * To be _trichotomous_ means that for any two `x` and `y`,
+  * EXACTLY one of the following three conditions holds:
+  *   - less(x, y)
+  *   - x == y
+  *   - less(y, x)
+  * Note that being trichotomous implies being irreflexive.
+  */
+  predicate Trichotomous<T(!new)>(less: (T, T) -> bool) {
+    (forall x, y :: less(x, y) || x == y || less(y, x)) &&  // at least one of the three
```
atomb, line 68: When I commented earlier about defining this for a comparison result data type, @cpitclaudel pointed out very rightly that my first attempt wouldn't work for comparison based on projection (e.g., looking only at the first element of a tuple), and I think this definition suffers from the same problem, using `==` here. That's not necessarily fatal, but it's something to be aware of.

src/Collections/Arrays/BinarySearch.dfy
```diff
@@ -0,0 +1,40 @@
...
+
+include "../../Wrappers.dfy"
+include "../../Relations.dfy"
+
+module BinarySearch {
+  import opened Wrappers
+  import opened Relations
+
+  method BinarySearch<T>(a: array<T>, key: T, less: (T, T) -> bool) returns (r: Option<nat>)
+    requires SortedBy(a[..], (x, y) => less(x, y) || x == y)
```
atomb, line 16: This will have the same issue I commented on for `Trichotomous`, making it impossible to sort by projections. Would it be possible to make the proof go through if we required `TotalOrdering` instead of `StrictTotalOrdering`?

src/Relations.dfy
atomb: Ah, but I notice this isn't actually used anywhere at this point. Should we remove it, given that issue?

atomb (approved):
This is a nice simplification of the original. :) Ordering continues to be a tricky topic, and I left a couple of comments related to it, though I'm not sure that any of them should hold up merging.

src/Collections/Sequences/MergeSort.dfy
atomb: It's perhaps worth noting that we could change this in the future without breaking client code. I'd maybe prefer `lessEq` or `lessThanOrEq`, to avoid the awkward reading of "equal than".

prvshah51: Thanks for suggestions , I am changing `compare` to `lessThanOrEq`.

MikaelMayer: > @MikaelMayer I just noticed you had some style requests that were lost in the shuffle. I'll respond to the ones I see that still seem relevant, but please just provide a quick comment at the top level if there are others we missed.
> 
> Re: the naming convention for predicates, e.g. "IsTotalOrder" instead of "TotalOrder", that's potentially a good idea but the latter convention is already established here in predicates like `Functions.Injective`. I'd suggest if you still feel strongly about it that you cut a repo issue for discussion, similar to #47.
> 
> Re: "lessThan" instead of "less", I don't have a strong opinion so I'd be willing to change it, personally. @prvshah51?

Could you use the attribute {:options "/functionSyntax:4"} and change all your predicate method to predicate ?

function method {:opaque} Sort<T>(s: seq<T>, compare: (T, T) -> bool) 
instead of a method

Add the following comment  to MergeSortBy
 // Splits a sequence in two, sorts the two subsequences using itself, and merge the two sorted sequences using `MergeSortedBy`

src/Collections/Arrays/BinarySearch.dfy
robin-aws: I attempted that at first, but I was discouraged not only by verification not immediately working, but also the fact that it makes the specification messier as well: instead of 

```dafny
ensures r.Some? ==> r.value < a.Length && a[r.value] == key
```

You would need

```dafny
ensures r.Some? ==> r.value < a.Length && lessThanOrEqual(a[r.value], key) && lessThanOrEqual(key, a[r.value])
```

Which is just plain much harder to read and understand.

I think for that reason, it makes more sense to improve this at the same time that we add comparison functions that return `CmpResult` instead as per #53.

src/Relations.dfy
robin-aws: Good point, I'd prefer to remove it too now that it's unused.

src/Collections/Sequences/MergeSort.dfy
robin-aws: > Could you use the attribute {:options "/functionSyntax:4"} and change all your predicate method to predicate ?
>
> function method {:opaque} Sort(s: seq, compare: (T, T) -> bool)
> instead of a method

I did that for `MergeSort.dfy`, but for `Relations.dfy` `SortedBy` was the only predicate method, and I actually made that a non-compiled `predicate` instead like the others. As per my comment on the original version in `dafny-reportgenerator`:

  // TODO: Make this a predicate-by-method so it has linear runtime instead of quadratic.
  // For now it's only used at runtime in tests.

This change doesn't need it to be compilable yet so I'd rather address that later.

> Add the following comment to MergeSortBy
> // Splits a sequence in two, sorts the two subsequences using itself, and merge the two sorted sequences using 
> MergeSortedBy

It's already there? :)

robin-aws: Trying to use `{:options "/functionSyntax:4"}` ran into what looks like a bug in Dafny 3.5.0 (which this repo is using) that I can't reproduce with the latest. I've reverted it for now and we can come back to that too: https://github.com/dafny-lang/libraries/issues/55

atomb (approved):

MikaelMayer (approved):
I approve but I still don't feel comfortable having a comparison function called twice in a binary search.

robin-aws:
> I approve but I still don't feel comfortable having a comparison function called twice in a binary search.

I definitely don't love it either. Both of these implementations are far from ideal when it comes to efficiency at runtime. I'm personally much more dissatisfied with my own MergeSort here rather than the binary search: it's using `s[..1]` expressions to recurse, and with the current compilation behavior that involves making a new copy, so the runtime is far worse than log-linear.

This whole repository needs more work to ensure it compiles to efficient-enough code for each supported language (see #24). In fact we're not even guaranteeing it is compilable at all for each target - it doesn't work for Java and hence the existence of https://github.com/dafny-lang/libraries-without-variance.

In general, I consider it good forwards progress to have a verified but inefficient definition of a concept here, since many users can live with the inefficiency in their context. But I'm looking forward to adding a lot of `by method`'s in the future! :)

PR #56, opened 2022-09-29
Make proofs easier for multiple Z3 versions

atomb:
Tested with these versions:

- 4.8.5
- 4.8.7
- 4.8.8
- 4.8.17
- 4.9.1
- 4.11.0
- 4.11.2

Two proofs fail with 4.8.7, and the whole process takes about twice as long with 4.11.0, but all other versions take around the same amount of time.

### DIFF
--- a/src/Collections/Sequences/LittleEndianNat.dfy
+++ b/src/Collections/Sequences/LittleEndianNat.dfy
@@ -181,7 +181,7 @@ abstract module LittleEndianNat {
 
   /* The nat representation of a sequence can be calculated using the nat
   representation of its prefix. */
-  lemma LemmaSeqPrefix(xs: seq<uint>, i: nat)
+  lemma {:vcs_split_on_every_assert} LemmaSeqPrefix(xs: seq<uint>, i: nat)
     requires 0 <= i <= |xs|
     ensures ToNatRight(xs[..i]) + ToNatRight(xs[i..]) * Pow(BASE(), i) == ToNatRight(xs)
   {
--- a/src/Collections/Sequences/LittleEndianNatConversions.dfy
+++ b/src/Collections/Sequences/LittleEndianNatConversions.dfy
@@ -58,8 +58,11 @@ abstract module LittleEndianNatConversions {
     ensures E > 0
   {
     LemmaDivBasicsAuto();
+    assert forall x :: x != 0 ==> 0 / x == 0;
     LemmaPowMultipliesAuto();
     LemmaFundamentalDivMod(Large.BITS(), Small.BITS());
+    assert Large.BITS() == Small.BITS() * (Large.BITS() / Small.BITS()) + (Large.BITS() % Small.BITS());
+    assert (Large.BITS() / Small.BITS()) > 0;
     Large.BITS() / Small.BITS()
   }
 
--- a/src/Collections/Sequences/Seq.dfy
+++ b/src/Collections/Sequences/Seq.dfy
@@ -559,6 +559,7 @@ module Seq {
     if |s| == 0 {
     } else {
       LemmaFlattenLengthLeMul(s[..|s|-1], j);
+      assert |FlattenReverse(s[..|s|-1])| <= (|s|-1) * j;
     }
   }
 
--- a/src/NonlinearArithmetic/Power.dfy
+++ b/src/NonlinearArithmetic/Power.dfy
@@ -199,6 +199,26 @@ module Power {
     }
   }
 
+  lemma LemmaPowSubs(b: int, e1: nat, e2: nat)
+    decreases e1
+    requires e1 >= e2
+    ensures Pow(b, e1 - e2) * Pow(b, e2) == Pow(b, e1)
+  {
+    LemmaPowAdds(b, e1 - e2, e2);
+  }
+
+  lemma LemmaPowSubsAuto()
+    ensures forall b: int, e1: nat, e2: nat {:trigger Pow(b, e1 - e2)} | e1 >= e2
+      :: Pow(b, e1 - e2) * Pow(b, e2) == Pow(b, e1)
+  {
+    reveal Pow();
+    forall b: int, e1: nat, e2: nat {:trigger Pow(b, e1 - e2)} | e1 >= e2
+      ensures Pow(b, e1 - e2) * Pow(b, e2) == Pow(b, e1)
+    {
+      LemmaPowSubs(b, e1, e2);
+    }
+  }
+
   /* Subtract exponents when dividing powers. */
   lemma LemmaPowSubtracts(b: nat, e1: nat, e2: nat)
     requires b > 0
@@ -209,7 +229,7 @@ module Power {
     LemmaPowPositiveAuto();
     calc {
       Pow(b, e2) / Pow(b, e1);
-        { LemmaPowAdds(b, e2 - e1, e1); }
+        { LemmaPowSubs(b, e2, e1); }
       Pow(b, e2 - e1) * Pow(b, e1) / Pow(b, e1);
         { LemmaDivByMultiple(Pow(b, e2 - e1), Pow(b, e1)); }
       Pow(b, e2 - e1);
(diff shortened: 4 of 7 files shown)


MikaelMayer (approved):
It's great that you can make the library work on more versions of Z3.
The code looks reasonable to me, even if I'm not a fan of putting `{:vcs_split_on_every_assert}` in a library, I think it might be the best decision for now.

atomb:
> It's great that you can make the library work on more versions of Z3. The code looks reasonable to me, even if I'm not a fan of putting `{:vcs_split_on_every_assert}` in a library, I think it might be the best decision for now.

Yeah, I wasn't thrilled by doing that, either. I hope that, at some point, we can work out a way to structure the proofs so that it's not necessary.

alex-chew (dismissed):
LGTM

src/Collections/Sequences/LittleEndianNat.dfy
```diff
@@ -181,7 +181,7 @@ abstract module LittleEndianNat {
 
   /* The nat representation of a sequence can be calculated using the nat
   representation of its prefix. */
-  lemma LemmaSeqPrefix(xs: seq<uint>, i: nat)
+  lemma {:vcs_split_on_every_assert} LemmaSeqPrefix(xs: seq<uint>, i: nat)
```
cpitclaudel, line 184: Is it OK to add `:vcs_split_on_every_assert` in the library?

src/Collections/Sequences/LittleEndianNatConversions.dfy
```diff
@@ -58,8 +58,11 @@ abstract module LittleEndianNatConversions {
     ensures E > 0
   {
     LemmaDivBasicsAuto();
+    assert forall x :: x != 0 ==> 0 / x == 0;
     LemmaPowMultipliesAuto();
     LemmaFundamentalDivMod(Large.BITS(), Small.BITS());
+    assert Large.BITS() == Small.BITS() * (Large.BITS() / Small.BITS()) + (Large.BITS() % Small.BITS());
+    assert (Large.BITS() / Small.BITS()) > 0;
```
cpitclaudel, line 65: Should these be lemma calls?

src/Collections/Sequences/Seq.dfy
```diff
@@ -559,6 +559,7 @@ module Seq {
     if |s| == 0 {
     } else {
       LemmaFlattenLengthLeMul(s[..|s|-1], j);
+      assert |FlattenReverse(s[..|s|-1])| <= (|s|-1) * j;
```
cpitclaudel, line 562: This assert just repeats the ensure, why is it needed?

src/NonlinearArithmetic/Power.dfy
```diff
@@ -199,6 +199,26 @@ module Power {
     }
   }
 
+  lemma LemmaPowSubs(b: int, e1: nat, e2: nat)
```
cpitclaudel, line 202: LemmaPowSubAddCancel maybe?

src/NonlinearArithmetic/Power.dfy
```diff
@@ -199,6 +199,26 @@ module Power {
...
+    LemmaPowAdds(b, e1 - e2, e2);
+  }
+
+  lemma LemmaPowSubsAuto()
+    ensures forall b: int, e1: nat, e2: nat {:trigger Pow(b, e1 - e2)} | e1 >= e2
+      :: Pow(b, e1 - e2) * Pow(b, e2) == Pow(b, e1)
+  {
+    reveal Pow();
+    forall b: int, e1: nat, e2: nat {:trigger Pow(b, e1 - e2)} | e1 >= e2
+      ensures Pow(b, e1 - e2) * Pow(b, e2) == Pow(b, e1)
```
cpitclaudel, line 216: You can remove this `ensures`

src/NonlinearArithmetic/Power.dfy
```diff
@@ -199,6 +199,26 @@ module Power {
...
+  {
+    LemmaPowAdds(b, e1 - e2, e2);
+  }
+
+  lemma LemmaPowSubsAuto()
+    ensures forall b: int, e1: nat, e2: nat {:trigger Pow(b, e1 - e2)} | e1 >= e2
+      :: Pow(b, e1 - e2) * Pow(b, e2) == Pow(b, e1)
+  {
+    reveal Pow();
+    forall b: int, e1: nat, e2: nat {:trigger Pow(b, e1 - e2)} | e1 >= e2
```
cpitclaudel, line 215: I think you can remove the trigger

src/Unicode/Utf8EncodingForm.dfy
```diff
@@ -249,9 +250,7 @@ module Utf8EncodingForm refines UnicodeEncodingForm {
     var z := (secondByte & 0xF) as bv24;
     var y := (thirdByte & 0x3F) as bv24;
     var x := (fourthByte & 0x3F) as bv24;
-    var v := (u1 << 18) | (u2 << 16) | (z << 12) | (y << 6) | x as Unicode.ScalarValue;
     assert {:split_here} true;
-    assert EncodeScalarValueQuadrupleByte(v) == m;
-    v
+    (u1 << 18) | (u2 << 16) | (z << 12) | (y << 6) | x as Unicode.ScalarValue
```
cpitclaudel, line 254: This change is a bit scary too: the performance improve if we remove the assert that restates the postcondition?

cpitclaudel (commented):
This is good to go as-is, but I left questions / suggestions

src/Collections/Sequences/LittleEndianNatConversions.dfy
atomb: They're specializations of the more general `ensures` clauses of `LemmaFundamentalDivMod`, so the only way I can think of to do that would be to have another fully-included but less general lemma to call here, which seems unsatisfying.

src/Collections/Sequences/LittleEndianNat.dfy
atomb: It's not ideal, IMO, but it's better than the status quo. I think it'd be better to make this proof easier in some other way, but it's not immediately clear how to do that. Any suggestions would be welcome!

src/Collections/Sequences/Seq.dfy
atomb: It's not exactly restating the postcondition: it's stating a more specific instance of it that also implies an equality. I think this general pattern, asserting an instantiation of a more general postcondition after a lemma call, is one of the steps we might need to start using more often to get proofs that verify more predictably.

src/Unicode/Utf8EncodingForm.dfy
```diff
@@ -232,6 +232,7 @@ module Utf8EncodingForm refines UnicodeEncodingForm {
     var z := (firstByte & 0xF) as bv24;
     var y := (secondByte & 0x3F) as bv24;
     var x := (thirdByte & 0x3F) as bv24;
+    assert {:split_here} true;
```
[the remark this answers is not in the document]
atomb: Can you say more specifically what you're worried about?

src/Unicode/Utf8EncodingForm.dfy
atomb: Apparently so! I'm not particularly worried that anything unsound is going on, if that's what you're worried about. I think the explanation is that Z3 is instantiating gobs of quantifiers, and in one case it gets lucky and in the other it doesn't.

src/NonlinearArithmetic/Power.dfy
atomb: Indeed I can! There are many other instances of this pattern in this file, but perhaps those could be addressed later. I'll simplify this one for now.

src/NonlinearArithmetic/Power.dfy
atomb: Yep! Same comment as for the `ensures`.

atomb:
> I would love to see in the CI a test that uses all versions of Z3.

Yeah, I'd like to see that at some point soon, too. The only thing stopping us from verifying the whole library with multiple versions is the compute time, and we could run them in parallel.

alex-chew (approved):

PR #58, opened 2022-10-21
Quick sort and related Arrays utilities

### LINKED ISSUE #54: Request: QuickSort (or any other more efficient alternative to MergeSort)
Follow-up from #49, which will close #39 but uses my quick-and-dirty-but-correct MergeSort from https://github.com/dafny-lang/dafny-reportgenerator/blob/main/src/StandardLibrary.dfy#L38. That code is actually very inefficient at runtime because it relies on `s[..1]` slices which currently impose a full copy.

I'd personally be happy with anything based on arrays instead of sequences, but as per #39 https://en.wikipedia.org/wiki/Timsort may be an even better option.

prvshah51:
Fix #54 


src/Collections/Arrays/Arrays.dfy
```diff
@@ -0,0 +1,374 @@
...
+include "../Sets/Sets.dfy"
+include "../../Comparison.dfy"
+
+module Arrays{
+
+import opened Sets
+import opened Comparison
+import opened Wrappers
+
+  trait Sorter<T> {
```
robin-aws, line 17: This definitely needs an example in `examples` to demonstrate using it. I assume you have to not only provide a concrete class that implements `Sorter<T>` and fills in `Cmp`, but perhaps even has to invoke a lemma or two from the trait in order to prove that it works.

Is it possible to also provide a `PredicateSorter` class that is constructed around a `(T, T) -> Cmp` function value? Perhaps it could even be a partial function if you also provide a ghost set of all values you intend to sort with it.

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT 
+*******************************************************************************/
+
+include "Wrappers.dfy"
+include "Relations.dfy"
+module Comparison {
+
+import opened Wrappers
+import opened Relations
```
robin-aws, line 13:
```suggestion
  import opened Wrappers
  import opened Relations
```

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+  datatype Cmp = Lt | Eq | Gt {
+    function method Flip(): Cmp {
+      match this
+        case Lt => Gt
+        case Eq => Eq
+        case Gt => Lt
+    }
+
+    const Le? := this != Gt
+    const Ge? := this != Lt
```
robin-aws, line 24: I definitely like this use of `?` at least, since it's a very natural extension of the built-in `Lt?/Eq?/Gt?` discriminator syntax.

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+
+    predicate Transitive?(ts: set<T>) {
+      forall t0, t1, t2 | t0 in ts && t1 in ts && t2 in ts :: Transitive??(t0, t1, t2)
+    }
+
+    predicate {:opaque} Valid?(ts: set<T>) {
+      Complete?(ts) && /* Antisymmetric?(ts) && */ Transitive?(ts)
+    }
+
+    predicate Sorted(sq: seq<T>) {
```
robin-aws, line 64: Why wouldn't this have a `?` as well then?

I'll be happiest if we come up with a simple rule for where `?` belongs. I definitely don't think it should be every predicate, especially since changing `Valid` to `Valid?` will be very disruptive for arguably not a lot of benefit. It looks like the pattern here is if the predicate is "about" the receiver rather than the arguments, but that feels very fuzzy to me, and I don't find it improves readability personally.

Perhaps to start, the straw-person proposal could be to only use it in symbols that aren't used like function calls, such as the `Le?` and `Ge?` constants above.

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+        case (Gt, Eq) => Some(Gt)
+        case (Gt, Gt) => Some(Gt)
+    }
+  }
+   datatype Comparison<!T> = Comparison(cmp: (T, T) -> Cmp) {
+    function method Compare(t0: T, t1: T): Cmp {
+      cmp(t0, t1)
+    }
+
+     predicate Complete??(t0: T, t1: T) {
```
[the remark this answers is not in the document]
robin-aws: Okay I see where the idea came from at least - you almost want to overload `Complete?` to accept either a pair or a set.

Could this be `CompleteOnPair` perhaps?

robin-aws (changes requested):
So naturally I've focussed most of my reviewing capacity on arguing about question marks. :) I do actually think it's a really important point for readability though, so I'd love to nail that down first before I look more deeply at the proofs.

src/Collections/Arrays/Arrays.dfy
```diff
@@ -0,0 +1,374 @@
...
+      ShuffledConcat(arr, lo, lo', hi'); ShuffledConcat(arr, lo, hi', hi);
+      assert Identical(arr, 0, lo) && Shuffled(arr, lo, hi) && Identical(arr, hi, arr.Length);
+      assert SameElements(arr, lo, hi);
+    }
+
+    twostate predicate Shuffled(arr: array<T>, lo: int, hi: int)
+      requires 0 <= lo <= hi <= arr.Length
+      reads arr
+    {
+      multiset(arr[lo..hi]) == old(multiset(arr[lo..hi]))
```
prvshah51, line 77: I remember adding this to almost every every function in `MergeSort.dfy` will need to replace that with `shuffled` instead.

src/Collections/Arrays/Arrays.dfy
```diff
@@ -0,0 +1,374 @@
...
+        SortedDutchFlag(arr, pivot, lo, left, right, hi);
+      }
+    }
+
+    function method MedianOfMedians(arr: array<T>, lo: int, hi: int): (t: T)
+      requires 0 <= lo < hi <= arr.Length
+      reads arr
+      ensures t in arr[lo..hi]
+    {
+      arr[lo] // TODO
```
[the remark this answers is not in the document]
cpitclaudel: Let's implement it — it's not hard and it's an important part of making the algorithm safe.

src/Comparison.dfy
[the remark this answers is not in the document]
cpitclaudel: Yeah, the `??` was due to have two layers.  Happy with calling them `Complete` and `Complete'` (the OCaml style) or `Complete1` (the Lisp style) or even `Complete_` or `Complete2`.  `CompletePair` is long.

src/Collections/Arrays/Arrays.dfy
robin-aws: Just to clarify, I'd strongly prefer to have an example, but the `PredicateSorter` idea can wait for a future PR.

src/Comparison.dfy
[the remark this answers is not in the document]
robin-aws: Okay, let's go with removing all of the single `?` suffixes (except for `Le?` and `Ge?` above), and replacing all of the double `??` suffixes with a `'` instead.

So in this case `Complete??` will become `Complete'`, and `Complete?` will become `Complete`.

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+
+    lemma SortedConcat(sq0: seq<T>, sq1: seq<T>)
+      requires Sorted(sq0)
+      requires Sorted(sq1)
+      requires forall i, j | 0 <= i < |sq0| && 0 <= j < |sq1| :: cmp(sq0[i], sq1[j]).Le?
+      ensures Sorted(sq0 + sq1)
+    {}
+
+    }
+   }
```
robin-aws, line 93:
```suggestion
  }
```

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+    predicate Complete?(ts: set<T>) {
+      forall t0, t1 | t0 in ts && t1 in ts :: Complete??(t0, t1)
+    }
+
+    predicate Transitive?(ts: set<T>) {
+      forall t0, t1, t2 | t0 in ts && t1 in ts && t2 in ts :: Transitive??(t0, t1, t2)
+    }
+
+    predicate {:opaque} Valid?(ts: set<T>) {
+      Complete?(ts) && /* Antisymmetric?(ts) && */ Transitive?(ts)
```
robin-aws, line 61:
```suggestion
      Complete?(ts) && */ Transitive?(ts)
```

Just following the general rule of not checking in commented out code. We could put a comment saying something like "we want to include `Antisymmetric(ts)` as well but it isn't necessary and makes verification timeout" (for example, I don't actually know in this case).

src/Comparison.dfy
```diff
@@ -0,0 +1,93 @@
...
+        case (Gt, Gt) => Some(Gt)
+    }
+  }
+   datatype Comparison<!T> = Comparison(cmp: (T, T) -> Cmp) {
+    function method Compare(t0: T, t1: T): Cmp {
+      cmp(t0, t1)
+    }
+
+     predicate Complete??(t0: T, t1: T) {
+      cmp(t0, t1) == cmp(t1, t0).Flip()
```
robin-aws, line 45: It threw me a bit that I didn't recognize this isn't one of the standard properties of total orderings, until I realized that "Transitive" really meant "cmp(...).Le? is a transitive binary relation" and similarly for "Reflexive". Is "complete" a standard term for this property of this kind of comparator?

I expect at the point where we try to connect `Relations` to this version, we'll have a lemma that proves that `C.Valid(s) ==> Relations.TotalOrdering((t1, t2) => C.cmp(t1, t2).Le?)`

src/Collections/Arrays/Arrays.dfy
```diff
@@ -0,0 +1,374 @@
...
+      assert arr[lo..mid] == arr[lo..hi][..mid-lo];
+    }
+
+    twostate lemma IdenticalShuffled(arr: array<T>, lo: int, hi: int)
+      requires 0 <= lo <= hi <= arr.Length
+      requires Identical(arr, lo, hi)
+      ensures Shuffled(arr, lo, hi)
+    {}
+
+    twostate predicate SameElements(arr: array<T>, lo: int, hi: int)
```
robin-aws, line 44: Consider `UnchangedExceptSliceShuffled` instead.

src/Collections/Arrays/Arrays.dfy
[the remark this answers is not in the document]
robin-aws: As per offline conversation we decided not to block on this, but in that case we should add a comment on QuickSort that it doesn't yet have the intended O(n log n) worst case runtime yet (which is what @cpitclaudel was referring to as "safe").

src/Comparison.dfy
robin-aws: See my other comment.

src/Collections/Arrays/Arrays.dfy
```diff
@@ -0,0 +1,374 @@
...
+        { assert old(arr[lo..hi] == arr[lo..mid] + arr[mid..hi]); }
+        old(multiset(arr[lo..mid] + arr[mid..hi]));
+        old(multiset(arr[lo..mid])) + old(multiset(arr[mid..hi]));
+        multiset(arr[lo..mid]) + multiset(arr[mid..hi]);
+        { assert arr[lo..hi] == arr[lo..mid] + arr[mid..hi]; }
+        multiset(arr[lo..hi]);
+      }
+    }
+
+    lemma Sortable_Slice(arr: array<T>, lo: int, hi: int, lo': int, hi': int)
```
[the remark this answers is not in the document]
prvshah51: `SortableSlice` or `SortableSubslice` both makes sense putting Subslice describes it better.

src/Collections/Arrays/Arrays.dfy
```diff
@@ -0,0 +1,375 @@
...
+import opened Sets
+import opened Comparison
+import opened Wrappers
+
+  trait Sorter<T> {
+    ghost const C := Comparison(Cmp)
+
+    function method Cmp(t0: T, t1: T): Cmp
+
+    twostate predicate UnchangedSlice(arr: array<T>, lo: int, hi: int)
```
robin-aws, line 22: Just food for thought, or a potential learning exercise: It might make all of this more readable if we had an actual `ArraySlice` datatype to pass around containing these values. It's a great example of the classic Design Pattern of introducing a type to hold collections of values that are always passed around together. Something like:

```dafny
  datatype ArraySlice_<T> = ArraySlice_(arr: array<T>, lo: int, hi: int) {
    predicate Valid() {
      0 <= lo <= hi <= arr.Length
    }

    twostate predicate Unchanged() 
      requires Valid()
      reads arr
    {
      arr[lo..hi] == old(arr[lo..hi])
    }

    // etc.
  }

  type ArraySlice<T> = a: ArraySlice_<T> | a.Valid() witness *
```

AFAICT this would be useful even if we only used it in ghost code to avoid any runtime cost (and even then it's possible we could optimize the wrapper away in the future, just as we now do for single-field datatypes).

src/Collections/Arrays/Arrays.dfy
robin-aws: Oh yikes, I went to write an example just to play around with it and immediately got the `A class may only extend a trait in the same module, unless the parent trait is annotated with {:termination false}` error. That means this is currently useless unless we add `{:termination false}`.

Unfortunately we should treat that as a hard blocker given there's already been resistance to putting `{:termination false}` on any standard library code: https://github.com/dafny-lang/libraries/pull/22 And especially given the discovery of https://github.com/dafny-lang/dafny/issues/2500 I'm not inclined to push back on that.

We can think about whether there's a good alternative for this code that doesn't use traits, but it might be better to stick with this if it's good UX and at least wait for the solution to https://github.com/dafny-lang/dafny/issues/2500.

robin-aws (changes requested):
I'm afraid I've realized this won't work as is and can't be made to work safely with current Dafny limitations - see my ~first~ [last](https://github.com/dafny-lang/libraries/pull/58#discussion_r1038594124) comment. We'll have to pause on this for a while.

src/Collections/Arrays/Arrays.dfy
cpitclaudel: That's not right: the proper API to use is `PredicateSorter`, which is a regular class, defined right below `Sorter`.  That doesn't require `:termination false`.

Original code here, with examples:  https://github.com/dafny-lang/compiler-bootstrap/blob/4f616822209828e48cf63d3da66ee1c010f689d4/src/Utils/Lib.Sort.dfy#L396-L420

PR #60, opened 2022-11-11
feat: add basic file I/O implementations

### LINKED ISSUE #50: Request: very basic file I/O
I've been intentionally cautious about this, since it requires linking to external code in each language no matter how you slice it. But I think the timing is right to experiment with very basic methods for reading and writing files, without streaming and without any model of the file system.

Since this repo contains only "pure" Dafny logic and no externs, I assert the I/O should go into a separate repo. That way we can continue to support consuming this repo as a simple submodule, whereas the I/O library will have to be packaged together with native code in some way.

alex-chew:
This PR adds basic file I/O facilities (read bytes from a file, write bytes to a file) to the for C#, Java, Javascript. Python support is nearly complete, but is blocked by what I believe is a missing external-module import in compiled code. Golang support is out of scope for now, since it's blocked by https://github.com/dafny-lang/dafny/issues/2989 and https://github.com/dafny-lang/dafny/issues/2045.

This PR was migrated from https://github.com/dafny-lang/dafny/pull/3018 for further development.

Fixes #50.

TODO:

- [x] Add doc comments to Dafny APIs
- [x] Add README explaining how to consume the module (other than the code examples)
- [x] Preserve exception stack traces
- [x] When writing, create parent directories if needed
- [x] Ensure paths work the same across target languages (and if possible, across platforms)
- ~~Fix Python support (currently unable to reference the newly defined `DafnyLibrary` from another compiled module)~~ (Python support to be added as a follow-up)


alex-chew:
For reviewers coming from the original PR (dafny-lang/dafny#3018) - the module/file structure in this PR is much simplified. Because we're no longer concerned with stuffing the internal file I/O implementations into existing runtime module names, we have the freedom to name the containing module in a way that works across all languages. So now including `FileIO.dfy` is all that's needed to reference the file I/O APIs from every target language - no target-specific includes/modules. During compile/run time, however, users will need to link the corresponding target-specific implementation (`FileIO.{cs,java,js,py}`) as it's not included in the runtime anymore.

This also means that consumers of the API no longer need to rely on module abstraction - just including and importing `FileIO` is sufficient. In this PR, for example, there are now only two test files (one for reading bytes and one for writing bytes).

davidcok:
Just checking that read from stdin and write to stdout will be possible with this PR

alex-chew:
> Just checking that read from stdin and write to stdout will be possible with this PR

This is not in scope for this PR.

alex-chew:
(CI is failing for unrelated reasons, but should pass once https://github.com/dafny-lang/libraries/pull/61 is merged.)

alex-chew:
The tests are failing because somehow the Dafny being run doesn't know about `--target` and a few other flags:

```
Dafny: Error: unknown switch: --no-verify
Dafny: Error: unknown switch: --target
Dafny: Error: unknown switch: --input
Dafny: Error: unknown switch: --
Use /help for available options

error: command failed with exit status: 1
```

Maybe `dafny run` isn't available in the version of Dafny we're using here?

MikaelMayer:
> Maybe `dafny run` isn't available in the version of Dafny we're using here?

Yes I think that's right. We did not update the dafny version in libraries for a long time.

.github/workflows/tests.yml
```diff
@@ -16,11 +16,14 @@ jobs:
       - name: Install Dafny
         uses: dafny-lang/setup-dafny-action@v1
         with:
-          dafny-version: "3.4.2"
+          dafny-version: "3.9.1"
```
robin-aws, line 19: I'd prefer that we don't do this until forced to. Ideally this code would be useable with the widest range of Dafny versions possible, especially since it's common for upgrades to be difficult due to changes in verification performance.

src/FileIO/FileIO.js
```diff
@@ -0,0 +1,68 @@
...
+   *
+   * We return these values individually because `Result` is not defined in the runtime but instead in library code.
+   * It is the responsibility of library code to construct an equivalent `Result` value.
+   */
+  $module.INTERNAL_ReadBytesFromFile = function(path) {
+    const emptySeq = _dafny.Seq.of();
+    try {
+      const readOpts = { encoding: null };  // read as buffer, not string
+      const buf = fs.readFileSync(path, readOpts);
+      const readBytes = _dafny.Seq.from(buf.valueOf(), byte => new BigNumber(byte));
```
robin-aws, line 29: Oof, we definitely have to optimize the crap out of the JS backend for byte sequences later :P

src/FileIO/README.md
```diff
@@ -0,0 +1,33 @@
+# File I/O
+
+The `FileIO` module provides basic file I/O operations.
+Right now, these are limited to reading bytes from a file, and writing bytes to a file.
+The API is intentionally limited in scope and will be expanded later.
+
+Unlike other modules in the `libraries` repo,
+the `FileIO` module will not compile or run correctly without a language-specific implementation file.
+Language-specific implementation files are provided for C#/.NET, Java, and Javascript.
+(Python and Golang support are planned.)
```
robin-aws, line 10: Looks like Python is in fact supported? :)

Was Go skipped for lack of time, or a deeper reason?

src/FileIO/README.md
```diff
@@ -0,0 +1,33 @@
...
+```bash
+# C#/.NET
+$ dafny run Program.dfy --include libraries/src/FileIO/FileIO.cs
+
+# Java
+$ dafny run Program.dfy --target:java --include libraries/src/FileIO/FileIO.java
+
+# Javascript
+$ dafny run Program.dfy --target:js --include libraries/src/FileIO/FileIO.js
+```
```
robin-aws, line 28: Missing the Python variant

examples/FileIO/ReadBytesFromFile.dfy
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+// RUN: echo -n > "%t"
```
robin-aws, line 7: Why is this needed?

examples/FileIO/ReadBytesFromFile.dfy
```diff
@@ -0,0 +1,45 @@
...
+// RUN: %baredafny run --no-verify --target:cs "%s" --input "%S/../../src/FileIO/FileIO.cs" -- "%S/data.txt" "System.ArgumentException:" >> "%t"
+// RUN: %baredafny run --no-verify --target:java "%s" --input "%S/../../src/FileIO/FileIO.java" -- "%S/data.txt" "java.io.IOException:" >> "%t"
+// RUN: %baredafny run --no-verify --target:js "%s" --input "%S/../../src/FileIO/FileIO.js" -- "%S/data.txt" "Error: ENOENT" >> "%t"
+// TODO: %baredafny run --no-verify --target:py "%s" --input "%S/../../src/FileIO/FileIO.py" -- "%S/data.txt" "[Errno 2]" >> "%t"
```
robin-aws, line 11: Wy doesn't this work?

examples/FileIO/WriteBytesToFile.dfy
```diff
@@ -0,0 +1,79 @@
...
+      {
+        var res := FileIO.WriteBytesToFile(outputDir + "/output_plain", bytes);
+        expect res.Success?, "unexpected failure writing to output_plain: " + res.error;
+      }
+      // Ensure that nonexistent parent directories are created as necessary
+      {
+        var res := FileIO.WriteBytesToFile(outputDir + "/foo/bar/output_nested", bytes);
+        expect res.Success?, "unexpected failure writing to output_nested: " + res.error;
+      }
+      // Ensure that file paths with .. are handled correctly
```
robin-aws, line 65: Worth mentioning that paths with `..` should always work in the top level API documentation?

robin-aws (changes requested):
Nice work, and well polished. I really appreciate the commitment to writing the same documentation in four different language dialects. :)

.github/workflows/tests.yml
robin-aws: Note that you can just change your LIT commands to use the old CLI syntax instead, e.g:

```
%dafny /compile:4 /noVerify /compileTarget:cs "%s" "%S/../../src/FileIO/FileIO.cs" /args "%S/data.txt" "System.ArgumentException:" >> "%t"
```

(Or close to that at least)

src/FileIO/README.md
alex-chew: Python is *almost* supported, but the Python compilation of the `FileIO` module doesn't correctly import `DafnyLibraries`. When I manually edit the `FileIO_Compile.py` to `import DafnyLibraries` at the top level, then everything works as expected. For that reason I've included the `FileIO.py` implementation here even though it can't quite be tested automatically.

src/FileIO/README.md
alex-chew: Python doesn't work right now, so I haven't added the corresponding command here.

examples/FileIO/ReadBytesFromFile.dfy
alex-chew: Same as https://github.com/dafny-lang/libraries/pull/60#discussion_r1038415542

examples/FileIO/ReadBytesFromFile.dfy
alex-chew: The `"%t"` temporary file must be truncated each time lit processes this file, or else subsequent `lit` invocations will lead to accumulating output there. (This really only applies locally, since we don't re-run tests in CI.)

* In `dafny-lang/dafny` we usually write `RUN: %dafny /compile:0 "%s" > "%t"` so we can diff against it, but that command's output isn't deterministic when run with the default `verificationLogger` parameters (which I don't see fit to disable here).
* We could change the first `RUN: %baredafny run ...` command to use `> "%t"` instead of `>> "%t"`, but I don't like having that difference between lines (in case someone reorders/removes lines, the truncation would break).
* So instead it seems easiest to just explicitly truncate the file with an `echo` (which lit handles as a [builtin command + redirection](https://github.com/llvm/llvm-project/blob/86fe4dfdb626d472f88bf8331da7cc98e9631cfe/llvm/utils/lit/lit/TestRunner.py#L317), anyway).

.github/workflows/tests.yml
alex-chew: As far as I can tell, v3.9.1 is the earliest version of Dafny under which these tests can reasonably run. That's because command-line arguments were added in v3.9.0, and they were [made consistent in JS](https://github.com/dafny-lang/dafny/pull/2876) in v3.9.1.

src/FileIO/README.md
alex-chew: Since we don't want to check in untested code, I've removed the Python bits.

examples/FileIO/WriteBytesToFile.dfy
alex-chew: Good call, fixed in latest commits.

.github/workflows/tests.yml
robin-aws: Got it, that's a harder motivation to work around.

This is fine for now then. If and when we get requests to guarantee this code will work on older versions of Dafny, we can test on a wider range. We may even need to do that on a per "library" basis.

examples/FileIO/ReadBytesFromFile.dfy
robin-aws: Ah interesting! I usually think of `%t` as a fresh temporary random value, which isn't true. In that case I agree this approach is a bit cleaner and more maintainable.

Although now that I look closer at this, I don't think there's any value in checking the output of these commands at all. All you care about is that the commands succeed. It's the same reason `%testDafnyForEachCompiler` over in `dafny-lang/dafny` doesn't look at the output of the initial verification call.

atomb: Yeah, I agree. The `.expect` file doesn't have any interesting content.

examples/FileIO/WriteBytesToFile.dfy.expect
```diff
@@ -0,0 +1,6 @@
+
```
atomb: This file also doesn't have particularly interesting content, so we could just check for successful execution.

examples/FileIO/ReadBytesFromFile.dfy
alex-chew: I suppose that's true, I've removed them.

src/FileIO/FileIO.cs
```diff
@@ -0,0 +1,94 @@
...
+    ///     </description>
+    ///   </item>
+    /// </list>
+    ///
+    /// We output these values individually because Result is not defined in the runtime but instead in library code.
+    /// It is the responsibility of library code to construct an equivalent Result value.
+    /// </summary>
+    public static void INTERNAL_ReadBytesFromFile(ISequence<char> path, out bool isError, out ISequence<byte> bytesRead,
+      out ISequence<char> errorMsg) {
+      isError = true;
```
atomb, line 41: I like that you initialize it this way. Starting out assuming an error and not indicating success until you've explicitly observed it is a nice practice.

examples/FileIO/WriteBytesToFile.dfy.expect
atomb: And while I was writing this you updated it. :)

robin-aws (approved):

src/FileIO/FileIO.dfy
```diff
@@ -0,0 +1,68 @@
...
+    var isError, errorMsg := INTERNAL_WriteBytesToFile(path, bytes);
+    return if isError then Failure(errorMsg) else Success(());
+  }
+
+  /*
+   * Private API - these are intentionally not exported from the module and should not be used elsewhere
+   */
+
+  method
+    {:extern "DafnyLibraries.FileIO", "INTERNAL_ReadBytesFromFile"}
```
atomb, line 60: Nice that you can use the same name for multiple languages!

atomb (approved):
This looks very nice!

src/FileIO/FileIO.dfy
alex-chew: Sadly this will require a bit more effort to make work in Python and Golang, but it certainly simplifies the module structure quite a bit compared to the original PR :)

PR #61, opened 2022-11-11
ci: avoid reinstalling dafny-reportgenerator

alex-chew:
As of [setup-dafny-action@v1.5.0](https://github.com/dafny-lang/setup-dafny-action/releases/tag/v1.5.0), dafny-reportgenerator is installed automatically, so our test workflow fails when it also tries to install the tool. This PR removes the installation step from the test workflow.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,9 +18,6 @@ jobs:
         with:
           dafny-version: "3.4.2"
 
-      - name: Install Dafny report generator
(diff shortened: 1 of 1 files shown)


atomb (approved):

MikaelMayer (approved):
Indeed, the new dafny actions are already installing the report generator, more specifically version 1.0.0
We could have run a command that check if it's already installed but for now that would be useless, so let's do as you suggested.

PR #64, opened 2023-01-03
First commit of the parsers combinator

MikaelMayer:

### DIFF
--- a/src/Parsers/README.md
+++ b/src/Parsers/README.md
@@ -1,0 +1,84 @@
+# Verified Parser Combinator Library in Dafny
+
+To view a full example of how to use the parser combinator library,
+especially how to define a recursive parser that is guaranteed to terminate,
+please refer to the file `exampleArithmetic.dfy`, which parses
+an arithmetic expression.
+
+To get started, first you need to import the parser, I recommend:
+
+```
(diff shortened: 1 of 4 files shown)


MikaelMayer:
Note of usability: 
- In the JSON example, extract the function `Token?(s: String)` whose definition is `WS().e_I(S(t)).??()`
- Add a comment to this function explaining what `WS()` and `??()` do

codyroux:
Some further minor notes:

- `StringParsersBuilders` should be `StringParserBuilders`, similarly for `ParsersBuilders`.
- Minor usability improvements: having an `alphanum` would be handy, and `Space` may want to exclude newlines (not sure). Also an `EnclosedBy<T>(left: string, p: B<T>, right: string)` or something.
- Some `ParsePair`, `ParseTriple`, `ParseQuadruple` would be nice. A little tedious to write, sadly.
- A `Token("foo", value)` implementation which simply does `S("foo").M( _ => value)` seems convenient.
- The implementation of `Digit` can be `CharTest(c => '0' <= c <= '9')` for efficiency, I think, unless there's some implementation specific which would make this incorrect. Or maybe one cannot prove anything about this definition.

None of these are "must haves" though.

PR #65, opened 2023-01-18
Improve durability of a few proofs

atomb:
LemmaNoDuplicatesInConcat would fail with certain random seeds, especially with newer Z3 versions. This change proves the result in smaller pieces rather than all in one go. It's arguably more readable, too.

LemmaFlattenLengthLeMul and LemmaFilterDistributesOverConcat also sometimes timed out or failed, but a few instances of `{:split_here}` seem to solve those issues.


fabiomadge (approved):

MikaelMayer (approved):
Indeed, the proof looks nicer. Great job !

PR #76, opened 2023-01-27
Adapt some proofs for lower variability and better compatibility with Z3 4.12.1

atomb:
This makes several proofs more explicit to reduce their variability and make them succeed with Z3 4.12.1.

### DIFF
--- a/examples/Collections/Arrays/BinarySearch.dfy
+++ b/examples/Collections/Arrays/BinarySearch.dfy
@@ -31,6 +31,7 @@ module BinarySearchExamples {
     var sortedInput := MergeSortBy(input, (x, y) => x <= y);
     print sortedInput, "\n";
 
+    assert SortedBy(sortedInput, (x, y) => x <= y);
     var sortedArray := ToArray(sortedInput);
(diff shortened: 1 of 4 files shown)


MikaelMayer (dismissed):
I haven't checked myself that the proof go better with more versions of z3 but the changes seem reasonable so I approve them. When can we expect to have some tests to ensure the library is tested with different versions of z3? I guess you still have some work to do, right?

atomb:
> I haven't checked myself that the proof go better with more versions of z3 but the changes seem reasonable so I approve them. When can we expect to have some tests to ensure the library is tested with different versions of z3? I guess you still have some work to do, right?

There might be a tiny bit of work left to get things working smoothly with multiple Z3 versions, but not much. Soon I'd like to update the CI to automatically verify with several versions (at least 4.8.5 and 4.12.1).

alex-chew (dismissed):

alex-chew (approved):

MikaelMayer (approved):
I haven't tested these changes but they look ok to me. Thanks for reducing variability.

PR #82, opened 2023-02-06
Unique minimum of totally-ordered sets

stefan-aws:
Establishes that every totally-ordered set has a unique minimum.

The proof is a simplified version of the arguments in https://leino.science/papers/krml275.html.

### DIFF
--- a/src/Relations.dfy
+++ b/src/Relations.dfy
@@ -76,5 +76,4 @@ module {:options "-functionSyntax:4"} Relations {
     requires TotalOrdering(lessThan)
     ensures SortedBy([x] + s, lessThan)
   {}  
-
 }
--- a/src/Relations.dfy
+++ b/src/Relations.dfy
@@ -49,6 +49,27 @@ module {:options "-functionSyntax:4"} Relations {
     forall i, j | 0 <= i < j < |a| :: lessThan(a[i], a[j])
   }
 
+  /* An element in an ordered set is called minimal, if it is less than every element of the set. */
+  ghost predicate IsMinimum<T>(R: (T, T) -> bool, m: T, s: set<T>) {
+    m in s && forall y: T | y in s :: R(m, y)
+  }
+
+  /* Any totally ordered set contains a unique minimal element. */
+  lemma LemmaUniqueMinimum<T(!new)>(R: (T, T) -> bool, s: set<T>) returns (m: T)
+    requires |s| > 0 && TotalOrdering(R)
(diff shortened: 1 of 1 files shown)


MikaelMayer (dismissed):
Straightforward, good job !

src/Relations.dfy
```diff
@@ -49,6 +49,27 @@ module {:options "-functionSyntax:4"} Relations {
...
 
+  /* An element in an ordered set is called minimal, if it is less than every element of the set. */
+  ghost predicate IsMinimum<T>(R: (T, T) -> bool, m: T, s: set<T>) {
+    m in s && forall y: T | y in s :: R(m, y)
+  }
+
+  /* Any totally ordered set contains a unique minimal element. */
+  lemma LemmaUniqueMinimum<T(!new)>(R: (T, T) -> bool, s: set<T>) returns (m: T)
+    requires |s| > 0 && TotalOrdering(R)
+    ensures IsMinimum(R, m, s) && (forall n: T :: IsMinimum(R, n, s) ==> m == n)
```
jtristan, line 60: I think that Rustan prefers that we use the syntactic sugar: 
```
forall n: T | IsMinimum(R, n, s) :: m == n
```

src/Relations.dfy
```diff
@@ -49,6 +49,27 @@ module {:options "-functionSyntax:4"} Relations {
...
+  {
+    var x :| x in s;
```
jtristan, line 62: I wonder if we should try to type variables explicitly? Is there some guideline on whether or not we want this for the library?

src/Relations.dfy
```diff
@@ -49,6 +49,27 @@ module {:options "-functionSyntax:4"} Relations {
...
+    if s == {x} {
+      m := x;
+    } else {
+      var m' := LemmaUniqueMinimum(R, s - {x});
+      if
```
jtristan, line 67: I don't really like that syntax so much, but it's just a matter of taste

jtristan (dismissed):

src/Relations.dfy
stefan-aws: The style guide http://dafny.org/dafny/StyleGuide/Style-Guide.html#type-declaration contains the guidance "If the type can be inferred by Dafny, leave it out, unless you think it provides useful documentation in the program.".

jtristan (approved):

MikaelMayer (approved):

PR #85, opened 2023-02-13
Draft user Guide to library

davidcok:


docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+the _Dafny Core Library_. Dafny programs are copmponed of modules; to ensure
+verification soundness, Dafny imposes a strict non-circular dependency relationship
+among modules. This Core library contains functionality that may be needed
+by other libraries, but it itself is standalone.
+
+All the supported, delivered libraries will
+- be stable --- a minimum of backwards incompatible changes
+- be verified --- proved to meet specifications and tested with expected use cases,
+with each release of the library and of the Dafny toolset
+- be efficient --- efficient in both proofs and executables that use the libraries
```
robin-aws, line 21: We should also have something about executable code being compilable for every supported target language by default, unless otherwise documented (such as the current state of FileIO)

docs/guide.md
```diff
@@ -0,0 +1,69 @@
+
+# Guide to the Dafny Core Library
+
+_THIS DOCUMENT IS A DRAFT UNDER ACTIVE DEBATE AND EDITING_
+
+A programming language library encapsulates a useful set of basic
+functionality that programmers can then incorporate in their custom
+program without having to rewrite, debug, or verify.
```
robin-aws, line 8: Feels obvious for the target audience, I'd omit personally but open to suggestion from other reviewers.

docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+- be efficient --- efficient in both proofs and executables that use the libraries
+
+The Dafny code for the library is written to be understandable with a minimum of inline documentation
+and is open source and available [on github](https://github.com/dafny-lang/libraries). This document is
+meant to be tutorial and example-based.
+
+## Content of the Core Library
+
+The sections below describe how to use each of the principal modules of the library:
+- [Dafny.Boxes](#sec-boxes) -- containers that enable encapsulaton of success and failure results in a single datatype
```
robin-aws, line 30: Not a bad alternate name. I wonder if we could add `AtomicBox` here too (although that would mean adding a target language dependency): https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntimeDafny/src/dafnyRuntime.dfy#L359

docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+- [Dafny.Math](#sec-math) --- basic mathematical operations and lemmas
+- [Dafny.Unicode](#sec-unicode) --- tools to manipulate unicode characters and strings
+- [Dafny.Collections](#sec-collections) -- functions and lemmas about sequences, sets and maps
+- [Dafny.NonlinearArith](sec-nonlinear-arith) --  lemmas and functions for non-linear arithmetic
+- [Dafny.Relations](#sec-relations) -- predicates and lemmas about properties of functions
+
+
+## How to use the library
+
+One uses the library on the command-line by including a path to the library in the `--library` option.
```
robin-aws, line 42: There will be extra steps for libraries with target language components like FileIO.

I was hoping we would have more of the features in https://github.com/dafny-lang/dafny/milestone/10 implemented to make this a smoother experience before we started labelling some of these as fully-published libraries.

docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+
+_THIS DOCUMENT IS A DRAFT UNDER ACTIVE DEBATE AND EDITING_
+
+A programming language library encapsulates a useful set of basic
+functionality that programmers can then incorporate in their custom
+program without having to rewrite, debug, or verify.
+
+Like other languages, the Dafny programming language and tool set includes
+libraries that provide functionality in various domains. This document describes 
+the _Dafny Core Library_. Dafny programs are copmponed of modules; to ensure
```
robin-aws, line 12: As I've mentioned offline, I'd like to aim for more, smaller libraries in general, so I'm nervous about committing to a single core library with all the code that's currently here or gets added in the future. But I think everything you've written applies to any library, so perhaps we can keep basically all of it but just say that we are currently considering it one big library, but may break it up in the future?

docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+---
+{% include_relative Math.md %}
+---
+{% include_relative Collections.md %}
+---
+{%include_relative Relations.md %}
+---
+{% include_realtive Unicode.md %}
+---
+{% include_relative NonlinearArithmetic.md %}
```
robin-aws, line 68: Could we move these to sit within their respective subdirectories instead?

robin-aws (commented):
Just added some top level thoughts for now, didn't yet review the individual library documentation pages yet.

docs/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+The `Dafny.BoundedInts` module contains definitions of types and constants for fixed-bit-width integers.
+
+Dafny itself generally uses and reasons about mathematical, unbounded integers and natural numbers.
+However compiled programs, for efficiency of computation and storage, will generally use fixed-bit-width
+numbers. So Dafny defines them here, for use when needed.
+
+An important implementation point about these type definitions is Dafny can determine which native type 
+in the target compiled program best matches the Dafny type. For example, the Dafny type `int16` is
+a signed integer that fits in 16 bits. If a program using this type is compiled to, say Java, then
+variables of this Dafny type will be compiled to Java `short` variables. In not every case is there
```
stefan-aws, line 13: "In not every case is there ..." sounds like contains a typo.

docs/Boxes.md
```diff
@@ -0,0 +1,140 @@
+
+## The `Boxes` module {#sec-boxes}
+
+A _Box_ datatype is one whose values can hold either a success indication, optionally along with a value, or a failure indication, optionally with error information.
```
stefan-aws, line 4: Your definition of a box datatype only matches a small subset of examples of monads/wrappers: essentially variations of the option/error monad. Other examples, such as the Giry monad, the Reader monad, the Either monad, etc. (currently not part of the library) will not be covered by this description.

Do you suggest that we have a separate module for the other examples? If that is the case, I propose we use a less generic name than "box" for this module, and instead choose something more specific which indicates the "error handling" nature of the examples.

docs/Boxes.md
```diff
@@ -0,0 +1,140 @@
...
+These are common programming idioms. The main complication comes when they are mixed in the same program.
+
+### Option
+
+Consider this routine that looks for a value in a sequence, beginning at position k, returning its index:
+```dafny
+import opened Dafny.Boxes
+function Find<T(==)>(s: seq<T>, k: int, value: T): (r: Option<int>)
+  requires 0 <= k <= |s|
+  ensures  r.Some? ==> s[r.???] == value;
```
stefan-aws, line 26: Is this a typo: `r.???` ?

stefan-aws, line 26:
```suggestion
  ensures  r.Some? ==> s[r.value] == value;
```

docs/Boxes.md
```diff
@@ -0,0 +1,140 @@
...
+  var index: int :- Find(s, 0, value);
+  // more calculations
+  return Some(value);
+}
+```
+
+You could just capture the result of find using `:=` in a `Option<int>` variable and inspect it. But if the `None` condition is 
+generally a rare error, it is easy to forget to always check that each operation was successful. Instead, the `:-` changes the 
+control flow so that if a `None` value is returned from `Find`, the method immediately aborts, with the output value (which has
+the `Option<int>` type) getting that returned value.So this operates something like exceptions.
```
stefan-aws, line 48:
```suggestion
the `Option<int>` type) getting that returned value. So this operates something like exceptions.
```

docs/Boxes.md
```diff
@@ -0,0 +1,140 @@
...
+method FindAllMatches<T(==)(s: seq<T>, value: T) returns (status: Outcome<string>, matches: seq<int>)
+```
+we can call it as
+```dafny
+var matches :- FindAllMatches(s, value);
+```
+
+Notice that there is no left-hand-side for the first out-parameter. It does not carry a value: if the value is a Fail, the control
+flow will abruptly return; if is is a Pass, the first out-parameter is then discarded and the second is assigned to the remaining LHS.
+An Outcome serves as an automatically-checked erorr mechanism.
```
stefan-aws, line 71:
```suggestion
An Outcome serves as an automatically-checked error mechanism.
```

docs/Boxes.md
```diff
@@ -0,0 +1,140 @@
...
+  // do various calculations
+  var index: int :- Find(s, 0, value).ToResult("not found");
+  // more calculations
+  return Success(value);
+}
+```
+
+Here we used the `ToResult` function of `Option` to convert the result value to a `Result`. There are similar `ToOutcome` and `ToOption` functions in each type.
+
+But what if we need to convert to some custom FC-type? Each datatype also has a `Map` function that takes a function as an argument, which function is a converter from
```
stefan-aws, line 111:
```suggestion
But what if we need to convert to some custom FC-type? Each datatype comes with a `Map` function, that can be given as an argument a function that converts from
```

stefan-aws: Not sure which `Map` function you are referring to, but it is not the usual one, and there is no alternative one in the current library.

Here you define an (anonymous) function `f: Option<int> -> Result<int, string>`, and this function can be applied to the output of type `Option<int>` of the function `Find`. I see no need to use `f` as an argument to a `Map` function, if you set up things in this way. Unless you define `Map(f) := f(this)`? If that's what you have in mind, I propose not to use the name 'Map', as this is often used to indicate the lifting of `f: A -> B` to e.g. `Option<A> -> Option<B>`.

docs/FileIO.md
```diff
@@ -0,0 +1,120 @@
+
+## The `FileIO` module {#sec-fileio}
+
+The `FileIO` module provides basic file I/O operations.
+Right now, these are limited to reading bytes from a file and writing bytes to a file.
+The API is intentionally limited in scope and will be expanded later.
+
+Unlike other modules in the `libraries` repo,
```
stefan-aws, line 8:
```suggestion
Unlike other modules in the `libraries` repository,
```

docs/Math.md
```diff
@@ -0,0 +1,8 @@
+
+## The `Math` module {#sec-math}
+
+The `Math` module contains just a few integer-related math functions (e.g. Min, Max, Abs).
```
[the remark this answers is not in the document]
stefan-aws: Makes you wonder if we need this module. Why not combine it with say `NonLinearArith`?

docs/guide.md
stefan-aws: Typo: _copmponed_ ?

docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+
+The Dafny code for the library is written to be understandable with a minimum of inline documentation
+and is open source and available [on github](https://github.com/dafny-lang/libraries). This document is
+meant to be tutorial and example-based.
+
+## Content of the Core Library
+
+The sections below describe how to use each of the principal modules of the library:
+- [Dafny.Boxes](#sec-boxes) -- containers that enable encapsulaton of success and failure results in a single datatype
+- [Bafny.BoundedInts](#sec-boundedints) -- definitions of types and cnostants for bounded integer arithmetic
```
stefan-aws, line 31:
```suggestion
- [Bafny.BoundedInts](#sec-boundedints) -- definitions of types and constants for bounded integer arithmetic
```

docs/guide.md
davidcok: Added a bullet stating this

docs/guide.md
davidcok: Removed. Boilerplate introduction.

docs/guide.md
davidcok: We can label it as draft-published, Dafny 4 ready code for now.

docs/guide.md
davidcok: Fixed the typo. Robin - I can't yet envision the workflow you envision for "lots of little libraries". Can you say a bit more about that?

docs/guide.md
davidcok: I'll try, but I'm quite sure that github pages does not support rendering pages outside the directory you point it to (in this case 'docs')

docs/Boxes.md
davidcok: A TBD - fixed.

davidcok: It is supposed to be Extract() -- fixed.

docs/Relations.md
```diff
@@ -0,0 +1,34 @@
...
+The `Relations` module defines a number of properties of relations (predicates of two arguments):
+- Reflexive
+- Irreflexive
+- AntiSymmetric
+- Transitive
+- Connected
+- StronglyConnected
+- TotalOrdering
+- StrictTotalOrdering
+And also this property of unary functions:
```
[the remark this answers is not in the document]
davidcok: Reworded

docs/Relations.md
```diff
@@ -0,0 +1,34 @@
...
+- Injective
+
+Some of these are needed for further reasoning. For example, to reason, for a function `f`,
+that `f(x) == f(y) ==> x == y`, one needs to prove that it is _Injective_. Similarly,
```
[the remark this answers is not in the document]
davidcok: My point was that if you have an arbitrary function and you want to be able to reason with the fact f(x) == f(y) ==> x == y. then it is useful to prove a general Injective lemma. But I left that example out.

docs/Boxes.md
davidcok: Do you have an alternate, concise, intuitive name?

Map(f) is indeed f(this) -- I added it because I wanted to provide the alternative of a streaming syntax x.Map(f) as well as f(x), for which no additional syntax is needed.

docs/Relations.md
```diff
@@ -1,7 +1,9 @@
 
 ## The `Relations` module {#sec-relations}
 
-The `Relations` module defines a number of properties of relations (predicates of two arguments):
+The `Relations` module defines a number of properties of functions.
```
stefan-aws, line 4:
```suggestion
The `Relations` module defines a number of properties of relations, which are predicates of two arguments. In particular, this module introduces properties of functions, which can be understood as relations of a particular type.
```

docs/Boxes.md
stefan-aws: Still thinking about an appropriate name.

docs/Relations.md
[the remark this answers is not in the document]
stefan-aws: I suggested a different wording.

docs/Boxes.md
```diff
@@ -0,0 +1,143 @@
...
+
+For this situation there is one more piece of syntax to make working with boxed values convenient. Suppose `Find` in our example is a method instead of a function. Then we have
+
+```dafny
+method m(s: seq<int>) returns (r: Result<int,string>) {
+  var value: int := *;
+  // do various calculations
+  var f = (o: Option<int>) => match o case Some(v) => Success(v) 
+                                      case None => Failure("not found")
+  var index: int :- {:convert f} Find(s, 0, value);
```
robin-aws, line 136: We don't currently have https://github.com/dafny-lang/dafny/issues/2436 planned for Dafny 4.0 so let's not mention any solution here.

docs/guide.md
davidcok: Got this reorg, though through regular links and not include_relative

docs/Boxes.md
davidcok: I'll leave it as tentatively planned.

docs/guide.md
robin-aws: I'm not in a good position to lay out the whole vision yet as I haven't had the time to refine it. :) I'll try to phrase my worries more concretely though.

To date, the `dafny-lang/libraries` (not "library"!) repo has only been consumed by using a submodule and `include`-ing whichever files you need. There hasn't been any concrete attempt to draw sensible boxes around individual "libraries". One of my worries is that if we go the easy route and call the whole thing one library, that will inevitably lead to the runtime footprint of a typical Dafny program getting larger and larger over time, and some target runtimes like Javascript care a lot about this footprint.

I also worry about making "the Core library" too special. If it's the only library we regression test as part of Dafny development, for example, it creates an evolutionary advantage for any code that lives here, and therefore any Dafny developer out there who wants to provide shared code will be incentivized to try to get it into the Core library, which makes Dafny maintainers a bottleneck. I'd be much happier to encourage a community and ecosystem for shared code to be created by anyone and good quality libraries to gain adoption naturally. It's also another case of "drinking our own champagne" if we develop libraries just like end users will.

All of this is ultimately just a nit about the "Dafny Core Library" label, because I don't think I want such a catch-all core library to exist long-term. But perhaps I'd be happy with just a better label for this repo as a collection of libraries.

docs/Boxes.md
robin-aws: If you are inclined to separate each type into its own module, OCaml-style, now is the time while you're reorganizing anyway. It will definitely be a fair pain for consumers, but better one big pain than multiple smaller pains that add up to more overall pain. :)

docs/guide.md
davidcok: Part of the meaning of "library" (singular or plural) is that you use what you need and not all the rest. Standard linking practice in producing an executable does just that -- only pulls in the material that the user (recursively) needs.
So we need to design the same structure for Dafny -- a means to lazily read and compile the source text as it is needed, not all at once at the beginning -- making sure that modules are no bigger than they need to be. That is a requirement whether the code is packaged as one big thing or many smaller things.

I'm happy to regression test anything we think is stable and wanted enough to expend the resources to watch it and fix it. I don't see this code as special other than that it is what we have.

As for others contributing -- well its open source and they are welcome to. We can state who the owner/maintainer is and when GitHub actions complain they can fix it.

I'm happy to adopt other intuitively-clear names. Just keep in mind that given Dafny's module dependency restrictions, some things are going to be leaves and others not.

docs/Boxes.md
davidcok: I would - but they depend on each other

docs/guide.md
robin-aws: Thinking more about this, I don't think "encapsulation of success and failure results in a single datatype" describes Option very well. It can be used like that as a result type, but it's usually better to provide more information about a "failure". It's used far more commonly for optional input, and hence the name.

My personal inclination is that "Wrappers" and "Boxes" are about equally "decent" rather than "amazing" so I'm not convinced the renaming cost is worth it yet. :)

docs/guide.md
robin-aws: What you're saying makes sense, but what worries me is assuming that every context that uses a single large library has the means to trim away what isn't used. Given Dafny compiles to many different source languages, and supports separate compilation, that doesn't seem like a safe assumption. Perhaps it's possible for Dafny itself to provide this feature eventually, but that seems like a more complicated solution than just delineating multiple smaller libraries in the first place.

To make it concrete, I was intending for us to just publish the content of `Wrappers.dfy` as a pre-compiled NuGet package, for applications like the Dafny ESDK to use. If we're calling this whole collection of code "the core library", that feel wrong and confusing.

davidcok: You are right that if there is no way to selectively "link" then the user has to do it manually. As a user I would find that a poor UX and annoying. But Dafny already constructs a module dependency graph, so omitting non-dependent modules from compilation does not seem like too much of a stretch. I could imagine even finer dependency tracking, but that would be more work.

Fine to publish Wrappers/Boxes as a start. Though to me it feels too small for the overhead.

robin-aws: > omitting non-dependent modules from compilation does not seem like too much of a stretch.

Ah yes, but that can only be done if you're compiling for a specific consumer, not if you're pre-compiling a library for arbitrary consumers.

If you're fine with the fact that we may publish subsets of "the core library" separately in the near future, whatever the meaning of "publish" is, I won't object to the naming for now.

docs/Boxes.md
```diff
@@ -0,0 +1,143 @@
...
+
+Any user datatype can serve this purpose, as long as it has an `IsFailure?` predicate 
+(in which case it is known as a [_failure-compatible_](https://dafny.org/latest/DafnyRef/DafyRef#TODO) type --- an FC-type). 
+In this module Dafny defines three such types, illustrates them with examples, and then describes 
+how they can be used in a system where different parts of the system use different FC-types.
+
+The three types are
+- `Option<R>` - which is either `Some` with a value of type `T` or a `None` with no information
+- `Outcome<E>` - which is either `Pass` with no informatino or `Fail` with an error value of type `E` (often a `string`) 
+- `Result<R,E>` - which is either `Success` with a value of type `R` or `Failure` with error value of type `E` (often a `string`)
```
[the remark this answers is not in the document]
seebees: I will also note that `Outcome` is primarily useful with `Need`.
Without this function it is of dubious use.
And I don't see any documentation on `Need` here,
which I would think of as a miss since it is primarily useful
when dealing with runtime checks for `Result` types.

docs/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
+
+## The `BoundedInts` module {#sec-boundedints}
+
+The `Dafny.BoundedInts` module contains definitions of types and constants for fixed-bit-width integers.
+
+Dafny itself generally uses and reasons about mathematical, unbounded integers and natural numbers.
+However compiled programs, for efficiency of computation and storage, will generally use fixed-bit-width
+numbers. So Dafny defines them here, for use when needed.
```
seebees, line 8: I would suggest that you use https://sembr.org/
This is because Markdown will compose the lines
into something readable by humans.
But when reading the written words,
the diffs get _much_ easier to see.

docs/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+- if you need a general unsigned 8-bit type, with values running up to 256, use `uint8`
+- if you want natural numbers that interact well with signed numbers and do not mind the restriction in range, use `nat8`
+- if the target platform for compilation does not have native unsigned int types, then use nat types because of the smaller range
+
+In addition, the module defines a number of constants that are powers of two (not all of them, just those that are generally useful).
+They are useful in stating the ranges of fixed-bit-width integers. Examples are `TWO_TO_THE_15`, `TWO_TO_THE_32`.
+
+Here are two example methods. In the first, the module `Dafny.BoundedInts` is renamed to `T`, which is then used as the prefix for type and constant names.
+In the second, the module is imported as opened, in which case the type and constant names can be used without qualification.
+<!-- %check-verify -->
```
seebees, line 33: very happy to see this!

docs/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
+
+## The `BoundedInts` module {#sec-boundedints}
+
+The `Dafny.BoundedInts` module contains definitions of types and constants for fixed-bit-width integers.
```
seebees, line 4: When dealing with bounded values
we also need subset types for dealing with `seq`.
This is because the minute that you touch a `seq`
or any of the other collections
you fall off the bounded value train.
So having a set of types for these things
is very powerful.

```dafny
predicate Int16?(i: int)
{
  i < T.TWO_TO_THE_15
}

type seq16<T> = s:seq<T> | Int16?(|s|)
```

docs/FileIO.md
```diff
@@ -0,0 +1,120 @@
...
+The API is intentionally limited in scope and will be expanded later.
+
+Unlike other modules in the `libraries` repository,
+the `FileIO` module will not compile or run correctly without a language-specific implementation file.
+Language-specific implementation files are provided for C#/.NET, Java, and Javascript.
+(Python and Golang support are planned.)
+Concretely, to use `FileIO` in your code, you must:
+
+1. `include` and `import` the `FileIO` module as you would any other library module
+2. incorporate the corresponding language-specific implementation file (e.g. `FileIO.cs`) when building or running your program
```
seebees, line 15: Should we touch on the problem of composing Dafny libraries?
e.g. if you try and use 2 Dafny libraries that both use `FileIO`
you are going to have a bad time.

docs/FileIO.md
```diff
@@ -0,0 +1,120 @@
...
+
+      var readBytes := seq(|res.value|, i requires 0 <= i < |res.value| => res.value[i] as int);
+      expect readBytes == expectedBytes, "read unexpected byte sequence";
+    }
+
+    // Failure path: attempting to read from a blank file path should never work.
+    {
+      var res := FileIO.ReadBytesFromFile("");
+      expect res.Failure?, "unexpected success";
+      expect expectedErrorPrefix <= res.error, "unexpected error message: " + res.error;
```
seebees, line 63: This is a good case for why strings are bad here.
Because a `FileNotFound` is not really an exceptional error case.
It should be easy to recover from.
But if you roll this library with string errors
it is complicated for customers to recover in this case.
This is especially hard in Dafny...

docs/Relations.md
```diff
@@ -0,0 +1,36 @@
...
+
+Properties of binary predicates:
+- Reflexive
+- Irreflexive
+- AntiSymmetric
+- Transitive
+- Connected
+- StronglyConnected
+- TotalOrdering
+- StrictTotalOrdering
```
seebees, line 14: This is wonderful.
But it would also be great if we had a few words on each.
Or a link to some more details :)

docs/guide.md
```diff
@@ -0,0 +1,69 @@
...
+All the supported, delivered libraries will
+- be stable --- a minimum of backwards incompatible changes
+- be verified --- proved to meet specifications and tested with expected use cases,
+with each release of the library and of the Dafny toolset
+- be efficient --- efficient in both proofs and executables that use the libraries
+- be compilable for every supported target language by default, with exceptions being documented and for exceptional reasons
+
+The Dafny code for the library is written to be understandable with a minimum of inline documentation
+and is open source and available [on github](https://github.com/dafny-lang/libraries). This document is
+meant to be tutorial and example-based.
```
seebees, line 22: For who?
There are many concepts in the Dafny ecosystem
that are well understood to long term Dafny users.
But to an average JS developer sometimes it is opaque :)

docs/guide.md
seebees: I agree that `Boxes` is, if any better, not enough better that I really want to refactor everything.
I don't think that a new user coming to Dafny is really helped here.

If you wanted a name that really helped I would suggest `Results` or `FailureCompatibleTypes` or `ElephantTypes`.

docs/guide.md
seebees: If you do this then you need to also include the published version in your target runtime project file.

docs/guide.md
davidcok: Is the Result datatype the most commonly used? If so there is some advantage to naming the module Result.

docs/Boxes.md
davidcok: Revised as suggested

docs/Boxes.md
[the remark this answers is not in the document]
davidcok: Included text and an example of Need

docs/BoundedInts.md
davidcok: Good

docs/BoundedInts.md
davidcok: I agree, but not yet in the scope of documentation!

docs/FileIO.md
davidcok: Yes. We have a fair bit of work to do to make a collection of useful code into one or more libraries. Tools need to be able to incorporate into the users compilation just as much from the library as is needed, no more, and no duplication. Something like a linker for Dafny.

docs/FileIO.md
davidcok: Makes an easy example. But I thoroughly agree - location information and something better than a string as an error id are helpful for good error catching/reporting/recovery in even a moderate-sized system.

docs/Relations.md
davidcok: I added some (mathematical detail)

davidcok:
Thanks for the comments.

docs/Relations.md
atomb: Have you pushed that yet? I don't see it, and I agree it'd be lovely to have.

docs/Relations.md
```diff
@@ -0,0 +1,36 @@
...
+  const IntLT  := ((i: int, j: int) => (i < j))
+```
+
+and then need to proof this lemma to use it in a sorting routine:
+```dafny
+  lemma IntLTisStrictTotalOrder()
+    ensures StrictTotalOrdering(IntLT) {}
+```
+
+Fortunately, dafny proofs this without aid.
```
atomb, line 34:
```suggestion
Fortunately, dafny proves this without aid.
```

atomb (commented):
Nice to see this shaping up! I just have some miscellaneous comments for now.

.github/workflows/tests.yml
```diff
@@ -1,10 +1,10 @@
-# This workflow performs static analysis checks.
+# This workflow does static verification of the legacy modules
 name: tests
 
 on:
   workflow_dispatch:
   push:
-    branches: [ master ]
+    branches: [ master , 'cok*' ]
```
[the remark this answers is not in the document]
davidcok: Yes - until we get through all this initial work.

docs/Boxes.md
```diff
@@ -0,0 +1,143 @@
...
+But what if we need to convert to some custom FC-type? Each datatype comes with a `Map` function, which can be given as an argument a function that converts from
+one FC-type to another. For example, we could rewrite the above as
+
+```dafny
+  var index: int :- Find(s, 0, value).
+     Map((o: Option<int>) => match o case Some(v) => Success(v) 
+                                     case None => Failure("not found") );
+```
+
+One can also just use usual function application syntax. `f(x)` to convert one FC-type
```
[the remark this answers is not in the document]
davidcok: Per a comment from Ryan, I changed to `Results`.
Revised and put in an example to avoid the confusion.

docs/Relations.md
davidcok: Fixed.

PR #86, opened 2023-02-13
Maps: add MapKeys

alex-chew:
This PR addresses a common use case where one wants to map an injective function over the keys of the map, retaining the values.

### DIFF
--- a/src/Collections/Maps/Maps.dfy
+++ b/src/Collections/Maps/Maps.dfy
@@ -1,18 +1,22 @@
 // RUN: %dafny /compile:0 "%s"
 
 /*******************************************************************************
-*  Original: Copyright 2018-2021 VMware, Inc., Microsoft Inc., Carnegie Mellon University, 
+*  Original: Copyright 2018-2021 VMware, Inc., Microsoft Inc., Carnegie Mellon University,
 *  ETH Zurich, and University of Washington
-*  SPDX-License-Identifier: BSD-2-Clause 
-* 
+*  SPDX-License-Identifier: BSD-2-Clause
+*
 *  Modifications and Extensions: Copyright by the contributors to the Dafny Project
-*  SPDX-License-Identifier: MIT 
+*  SPDX-License-Identifier: MIT
 *******************************************************************************/
 
 include "../../Wrappers.dfy"
+include "../../Functions.dfy"
+include "../Sets/Sets.dfy"
 
 module {:options "-functionSyntax:4"} Maps {
   import opened Wrappers
+  import Functions
+  import Sets
 
   function Get<X, Y>(m: map<X, Y>, x: X): Option<Y>
   {
@@ -127,4 +131,16 @@ module {:options "-functionSyntax:4"} Maps {
     forall x, x' {:trigger m[x], m[x']} :: x in m && x' in m && start <= x <= x' ==> m[x] <= m[x']
   }
 
+  /* Map an injective function over the keys of a map, retaining the values. */
+  function {:opaque} MapKeys<X(!new), Y, X'>(m: map<X, Y>, f: X --> X'): (m': map<X', Y>)
+    reads f.reads
+    requires forall x {:trigger f.requires(x)} :: f.requires(x)
+    requires Functions.Injective(f)
+    ensures |m'| == |m|
+    ensures m'.Values == m.Values
+   {
+    var m' := map k <- m :: f(k) := m[k];
+    Sets.LemmaMapSize(m.Keys, m'.Keys, f);
+    m'
+  }
 }


davidcok (approved):

src/Collections/Maps/Maps.dfy
```diff
@@ -127,4 +131,16 @@ module {:options "-functionSyntax:4"} Maps {
     forall x, x' {:trigger m[x], m[x']} :: x in m && x' in m && start <= x <= x' ==> m[x] <= m[x']
   }
 
+  /* Map an injective function over the keys of a map, retaining the values. */
```
MikaelMayer, line 134:
```suggestion
  /* Maps an injective function over the keys of a map, retaining the values. */
```

MikaelMayer (commented):
Minor typo, otherwise looks good!

src/Collections/Maps/Maps.dfy
```diff
@@ -127,4 +131,16 @@ module {:options "-functionSyntax:4"} Maps {
...
+    requires forall x {:trigger f.requires(x)} :: f.requires(x)
+    requires Functions.Injective(f)
+    ensures |m'| == |m|
+    ensures m'.Values == m.Values
+  {
+    var m' := map k <- m :: f(k) := m[k];
+    Sets.LemmaMapSize(m.Keys, m'.Keys, f);
+    m'
+  }
 }
```
davidcok, line 146: Please add an example that uses the function and proves interesting things about its effects -- to be sure the proofs are successful in reasonable time.

davidcok (changes requested):

src/Collections/Maps/Maps.dfy
```diff
@@ -127,4 +131,16 @@ module {:options "-functionSyntax:4"} Maps {
     forall x, x' {:trigger m[x], m[x']} :: x in m && x' in m && start <= x <= x' ==> m[x] <= m[x']
   }
 
+  /* Maps an injective function over the keys of a map, retaining the values. */
+  function {:opaque} MapKeys<X(!new), Y, X'>(m: map<X, Y>, f: X --> X'): (m': map<X', Y>)
+    reads f.reads
+    requires forall x {:trigger f.requires(x)} :: f.requires(x)
```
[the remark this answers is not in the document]
alex-chew: `forall x :: f.requires(x)` is a precondition of `Functions.Injective(f)`, so it needs to be propagated here.

Alternatively, one can define a restricted form of `Functions.Injective` like so:
```dafny
  ghost predicate InjectiveRestricted<X(!new), Y>(f: X --> Y, restriction: set<X>)
    reads f.reads
    requires forall x | x in restriction :: f.requires(x)
    ensures Injective.requires(f) && Injective(f) ==> InjectiveRestricted(f, restriction)
  {
    forall x1 <- restriction, x2 <- restriction :: f(x1) == f(x2) ==> x1 == x2
  }
```
and indeed the above verifies. But when trying to use this in `Sets.LemmaMapSize` I ran into proof issues with its other preconditions, and I decided not to pursue the idea further.

src/Collections/Maps/Maps.dfy
```diff
@@ -127,4 +131,16 @@ module {:options "-functionSyntax:4"} Maps {
...
+    requires Functions.Injective(f)
+    ensures |m'| == |m|
+    ensures m'.Values == m.Values
```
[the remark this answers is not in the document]
alex-chew: This seems useful and requires no additional proof, so I'll go ahead and add it.

src/Collections/Maps/Maps.dfy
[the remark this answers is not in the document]
davidcok: OK

src/Collections/Maps/Maps.dfy
```diff
@@ -93,6 +97,39 @@ module {:options "-functionSyntax:4"} Maps {
     forall x, x' {:trigger m[x], m[x']} :: x != x' && x in m && x' in m ==> m[x] != m[x']
   }
 
+  ghost predicate {:opaque} Contains<X, Y>(big: map<X, Y>, small: map<X, Y>)
```
robin-aws, line 100: Consider `SubmapOf` (and reversing the arguments) instead, bit more consistent with the other `Collections` code.

Bit unfortunate Dafny doesn't define `<=` for maps. :)

src/Collections/Maps/Maps.dfy
```diff
@@ -127,4 +164,33 @@ module {:options "-functionSyntax:4"} Maps {
     forall x, x' {:trigger m[x], m[x']} :: x in m && x' in m && start <= x <= x' ==> m[x] <= m[x']
   }
 
+  /* Maps an injective function over the keys of a map, retaining the values. */
+  function {:opaque} MapKeys<X(!new), Y, X'>(m: map<X, Y>, f: X --> X'): (m': map<X', Y>)
```
robin-aws, line 168: Unfortunate that `Seq.Map` has the parameters in the other order. This is something we should align on better in the `src/dafny/...` copy of all the collections code.

robin-aws (approved):
LGTM, made some non-blocking suggestions (because we have the opportunity to improve naming before this code is more locked down)

robin-aws:
See also: https://github.com/dafny-lang/libraries/issues/119. I don't want to block on adding this under `src/dafny` as well but do want to track that work.

src/Collections/Maps/Maps.dfy
alex-chew: Turns out `IsSubset` was already defined, so I've switched everything to use that and removed the redundant `Contains`.

src/Collections/Maps/Maps.dfy
alex-chew: Since these are new APIs I'll just switch them around now.

alex-chew:
> See also: https://github.com/dafny-lang/libraries/issues/119. I don't want to block on adding this under src/dafny as well but do want to track that work.

I'll take care of syncing to `src/dafny` as a follow-up, once this PR is finalized.

robin-aws (approved):

PR #87, opened 2023-02-15
Mutable maps via Java extern

stefan-aws:
Per request, adds support for mutable maps by interfacing with Java. Partly follows https://github.com/dafny-lang/dafny/wiki/Modeling-External-State-Correctly and https://github.com/dafny-lang/libraries/tree/master/src/FileIO.


src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
...
+
+    method {:extern "put"} Put(k: K, v: V)
+      modifies this
+      ensures this.content() == old(this.content())[k := v]    
+
+    function {:extern "keys"} Keys(): (keys: set<K>)
+      reads this
+      ensures keys == this.content().Keys
+
+    function {:extern "values"} Values(): (values: set<V>)
```
robin-aws, line 24: Could we leave off `Values()` and `Items()` for now, so that we don't force `V` to be equality-supporting unnecessarily?

Alternatively, you could make those methods instead of functions and have them return non-deterministically ordered sequences instead of sets.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
...
+      reads this
+      ensures values == this.content().Values
+
+    function {:extern "items"} Items(): (items: set<(K,V)>)
+      reads this
+      ensures items == this.content().Items
+      ensures items == set k | k in this.content().Keys :: (k, this.content()[k])
+
+    function {:extern "find"} Find(k: K): (v: V)
```
robin-aws, line 33: Not the most idiiomatic name IMO, I think of "find" as being a search through a datatype that isn't build for indexing, like a list. `m[d]` is called "map selection" in the reference manual so I'd prefer "Select".

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+
+module {:extern "MutableMap"} {:options "/functionSyntax:4"} MutableMap {
```
robin-aws, line 8: I'd at least move this under the "DafnyLibraries" namespace as the FileIO code is. "MutableMap" has an outside chance of conflicting with other Java code.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
...
+  class {:extern "MutableMap"} MutableMap<K(==),V(==)> {
+    function {:extern "content"} content(): map<K, V>
+      reads this
+
+    constructor {:extern "MutableMap"} ()
+      ensures this.content() == map[]
```
robin-aws, line 14: Generally better to provide a static factory method as part of your API instead of a constructor, so the concrete type can be changed without breaking anyone.

This suggests making `MutableMap` a trait rather than a class is a good idea, which will also help with the feasibility implementation as per my comment there.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
...
+      ensures keys == this.content().Keys
+
+    function {:extern "values"} Values(): (values: set<V>)
+      reads this
+      ensures values == this.content().Values
+
+    function {:extern "items"} Items(): (items: set<(K,V)>)
+      reads this
+      ensures items == this.content().Items
+      ensures items == set k | k in this.content().Keys :: (k, this.content()[k])
```
robin-aws, line 31: This is technically redundant, did you find it necessary? It feels like it should be a separate lemma from Maps.dfy instead.

src/MutableMap/MutableMapFeasability.dfy
```diff
@@ -0,0 +1,126 @@
...
+
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+
+module {:options "/functionSyntax:4"} MutableMapTrait {
+  trait MutableMapTrait<K(==),V(==)> {
```
robin-aws, line 11: Love the feasibility implementation! There are ways of avoiding duplicating the specification without resorting to `{:termination false}` though. For example see what I did for the bootstrapping `Sequence<T>` trait here: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyRuntime/DafnyRuntimeDafny/src/dafnyRuntime.dfy Namely, I have the actual trait in an abstract module that is refined per target language, and that means I can refine for Dafny as well.

src/MutableMap/MutableMap.java
```diff
@@ -0,0 +1,63 @@
...
+import dafny.Tuple2;
+import dafny.TypeDescriptor;
+
+import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.Map;
+
+
+public class MutableMap<K,V> {
+  private HashMap<K, V> m;
```
robin-aws, line 18: `ConcurrentHashMap<K, V>` instead please. :) Dafny can't prove correctness under concurrent execution in general, but we still want our runtimes and libraries to be threadsafe enough that it is at least possible to compile to threadsafe code.

src/MutableMap/MutableMap.dfy
robin-aws: I'm especially inclined to leave these out given https://github.com/dafny-lang/dafny/issues/1373

robin-aws (changes requested):
Looks great! Lots of suggestions for polish but nothing serious.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
...
+
+    function {:extern "find"} Find(k: K): (v: V)
+      reads this
+      requires k in this.Keys()
```
alex-chew, line 35: While requiring `k in this.Keys()` makes some use cases easier, there's a performance pitfall if user code wants to "get the value associated with `k` or something else if `k` isn't in the map". A user might write e.g.
```dafny
let v = if k in mutableMap.Keys() then Some(mutableMap.Find(k)) else None;
```
in which case the `.Keys()` call creates a new `DafnySet` via a `HashSet` constructor that individually copies each element. The better alternative is to write
```dafny
let v = if k in mutableMap.content() then Some(mutableMap.Find(k)) else None;
```
which creates a new `DafnyMap` but does not copy any elements.

I think it'd be best to add a `function FindIfPresent(k: K): (v: Option<V>)` or something along those lines, both to capture this common use case and to help steer users away from implementing it themselves inefficiently.

(For the same reasons, you might consider adding a `predicate HasKey(k: K)` implemented by each runtime. Then `FindIfPresent` can just combine `HasKey` and the existing presence-requiring `Find`.)

examples/MutableMap/MutableMapExamples.dfy
```diff
@@ -0,0 +1,43 @@
...
+// RUN: %dafny /compile:0 "%s"
+
+include "../../src/MutableMap/MutableMap.dfy"
+
+module MutableMapExamples {
+  import opened MutableMap
+
+  method Main() {
+    var m := new MutableMap<string,string>();
+    assert m.Keys() == {};
```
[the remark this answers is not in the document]
davidcok: I fully echo that there should be an executable example also, that tests all the calls and corner cases. Ideally it is the same method with both asserts and expects. Also should have example programs that test that preconditions are checked, along with testing that postconditions are strong enough to reason about.

davidcok (dismissed):
I echo much of what Robin wrote, but did not pile on.

src/MutableMap/MutableMap.dfy
stefan-aws: At some point it seemed like I need both ensures clauses to prove the assert statements in `MutableMapExamples.dfy`. But now they go through with just the second. I will remove the first one.

src/MutableMap/MutableMap.dfy
stefan-aws: It's called `Find` in OCaml (https://v2.ocaml.org/api/Map.Make.html), that's why I went with it. But I'm fine with `Select`, too.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,47 @@
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+
+module {:extern "MutableMap"} {:options "/functionSyntax:4"} MutableMap {
+  class {:extern "MutableMap"} MutableMap<K(==),V(==)> {
```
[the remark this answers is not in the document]
stefan-aws: This leads to the error `class declarations only support non-variant type parameters`.

src/MutableMap/MutableMap.dfy
davidcok: Even if an ensures clause is not needed to make the example proofs go through, I would tend to keep it for a library if it seemed it might save the prover some work or make a fact know that a user proof might need. The example program has lots of asserts so it might inadvertently guide the prover where a a less sparsely annotated program might not. I'd be good with keeping it. Your call.

src/MutableMap/MutableMap.dfy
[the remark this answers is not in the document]
davidcok: OK. I'll look into that - are they only applicable to datatypes?

src/MutableMap/MutableMapFeasability.dfy
atomb: After an in-depth offline investigation, @stefan-aws and I decided that leaving the contract redundancy in for now is okay. Because the test case in this example aims to test the Java code from Dafny, the approach used in the Go runtime cited above doesn't quite work directly. There may be a way to do something similar to reduce the redundancy in the future, but we should be able to do so without changing the public API.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,56 @@
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+
+include "MutableMapTrait.dfy"
+
+module {:extern "DafnyLibraries"} {:options "-functionSyntax:4"} MutableMap refines MutableMapTrait {
```
robin-aws, line 10: Having a real mismatch between the Dafny and extern names for a module seems like trouble, can we nest it in Dafny as `DafnyLibraries.MutableMap` instead?

examples/MutableMap/MutableMapExamples.dfy
```diff
@@ -0,0 +1,45 @@
...
+module {:options "-functionSyntax:4"} MutableMapExamples {
+  import opened MutableMap
+
+  method Main() {
+    var m := new MutableMap<string,string>();
+    assert m.Keys() == {};
+    assert m.Values() == {};
+    assert m.Items() == {};
+    assert m.Size() == 0;
+    m.Put("testkey", "testvalue");
```
robin-aws, line 19: readability nit: a newline before each (operation + asserts) block would be really nice. :)

src/MutableMap/MutableMap.java
```diff
@@ -0,0 +1,70 @@
...
+  public DafnySet<? extends V> Values() {
+    return new DafnySet(m.values());
+  }
+
+  public DafnySet<? extends Tuple2<K,V>> Items() {
+    ArrayList<Tuple2<K, V>> list = new ArrayList<Tuple2<K, V>>();
+    for (Map.Entry<K, V> entry : m.entrySet()) {
+      list.add(new Tuple2<K, V>(entry.getKey(), entry.getValue()));
+    }
+    return new DafnySet<Tuple2<K, V>>(list);
```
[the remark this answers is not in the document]
robin-aws: As in `return new DafnySet(m.entrySet())`?

src/MutableMap/MutableMapDafny.dfy
```diff
@@ -0,0 +1,125 @@
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+
+include "MutableMapTrait.dfy"
+   
+module {:options "-functionSyntax:4"} MutableMapDafny refines MutableMapTrait {
```
robin-aws, line 10: Add a comment that this is a feasibility implementation. Probably better to move both this and the trait version to `examples`, since they are like tests of the source functionality but not meant to be used by library consumers.

src/MutableMap/MutableMapTrait.dfy
```diff
@@ -0,0 +1,66 @@
...
+*******************************************************************************/
+
+// RUN: %dafny /compile:0 "%s"
+
+include "../Wrappers.dfy"
+
+module {:options "-functionSyntax:4"} MutableMapTrait {
+  import opened Wrappers
+
+  trait {:termination false} MutableMapTrait<K(==),V(==)> {
```
robin-aws, line 13: Add a comment that this is duplicated here for the benefit of the feasibility implementation.

examples/MutableMap/MutableMapExamples.dfy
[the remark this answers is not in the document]
robin-aws: Unresolving this to track the fact that my suggestion above led to compilation errors, which we're going to figure out in case they are a real issue with using the library.

src/MutableMap/MutableMap.java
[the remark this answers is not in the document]
alex-chew: I'll resolve this conversation as it's not applicable, but [`DafnyMap#entrySet`](https://github.com/dafny-lang/dafny/blob/e956ba4322e645f70f5e215c379ff78d50ae11b0/Source/DafnyRuntime/DafnyRuntimeJava/src/main/java/dafny/DafnyMap.java#L153) has a nearly identical implementation so I thought there was a way to avoid the duplicate code with `return DafnyMap(m).entrySet()`. Turns out that wasn't possible because the `DafnyMap(HashMap)` constructor is private; and it isn't possible now because `m` is a `ConcurrentHashMap`.

src/MutableMap/MutableMap.dfy
stefan-aws: On the Java side, we have a class `MutableMap` within a package `DafnyLibraries`. I am now following the same pattern and naming in Dafny. I don't think having two nested modules within Dafny works, if the Java side only has one module.

stefan-aws: Actually, importing in Dafny via `import opened DafnyLibraries` seems odd. So either: 
i) we find a way to use nested modules (which I don't think works with `:extern`, since we only have one on the Java side)
ii) we stick to the original solution, which used the name `MutableMap` in Dafny, and linked it to `DafnyLibraries` in Java
iii) we change the package name `DafnyLibraries` to `MutableMap` on the Java side

examples/MutableMap/MutableMapExamples.dfy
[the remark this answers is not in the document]
stefan-aws: Added `AssertAndExpect` and resolved compilation errors

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,75 @@
+/*******************************************************************************
+*  Copyright by the contributors to the Dafny Project
+*  SPDX-License-Identifier: MIT
+*******************************************************************************/
+
+// RUN: %verify "%s"
+
+include "../../examples/Mutablemap/MutableMapTrait.dfy"
```
robin-aws, line 8:
```suggestion
include "../../examples/MutableMap/MutableMapTrait.dfy"
```

src/MutableMap/MutableMap.dfy
stefan-aws: Discussed this with @atomb, didn't work out as hoped. I suggest we leave it as it is and potentially address this again in the future.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,75 @@
...
+// RUN: %verify "%s"
+
+include "../../examples/MutableMap/MutableMapTrait.dfy"
+include "../../src/Wrappers.dfy"
+
+/**
+  *  Implements mutable maps by interfacing with external code, e.g. "MutableMap.java".
+  */
+
+module {:extern "DafnyLibraries"} {:options "-functionSyntax:4"} MutableMap {
```
robin-aws, line 15:
```suggestion
module {:extern "DafnyLibraries"} {:options "-functionSyntax:4"} DafnyLibraries {
```

There's no benefit to using such a different name in Dafny, only confusion

examples/MutableMap/MutableMapTrait.dfy
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+  *  Specifications that should be satisfied by any implementation of mutable maps.
+  *  Possible instantiations are given in "MutableMapDafny.dfy" (not meant for usage, 
+  *  only exists to verify feasability) and "../../src/MutableMap/MutableMap.dfy" 
+  *  (meant for usage; interfaces with external code, e.g. "../../src/MutableMap/
+  *  MutableMap.java").
+  */
+module {:options "-functionSyntax:4"} MutableMapTrait {
+  trait {:termination false} MutableMapTrait<K(==),V(==)> {
```
robin-aws, line 16: Unfortunately I think you'd better duplicate this trait in both `src` and `examples`, at least until dafny-lang/dafny#2500 is resolved.

robin-aws (changes requested):
Looks great, just a couple of bookkeeping changes to do

robin-aws:
Oh one other thing for later: users might expect more to find this under "Collections". We may want to consider that when we make a copy for this under DafnyCore as well.

src/MutableMap/MutableMap.dfy
```diff
@@ -0,0 +1,120 @@
...
+/**
+  *  Implements mutable maps by interfacing with external code, e.g. "MutableMap.java".
+  */
+
+module {:extern "DafnyLibraries"} {:options "-functionSyntax:4"} DafnyLibraries {
+  import opened Wrappers
+
+  /**
+  *  NOTE: Only here because of #2500; once resolved import "../../examples/MutableMap/
+  *  MutableMapTrait.dfy".
```
robin-aws, line 19: nit for later: it seems useful to have the trait for consumers too, so my preference would be to remove the example copy instead. But this is fine for now.

robin-aws (approved):
Hooray for 🔇 🅰️ 🔔  🗺️s!

alex-chew (approved):

PR #89, opened 2023-02-16
Fix "--include" typo

stefan-aws:
The reference [manual](http://dafny.org/dafny/DafnyRef/DafnyRef) doesn't include an `--include` option. I believe this should be `--input`.


alex-chew (approved):
Oops, you're absolutely right. Good catch!

keyboardDrummer (approved):

Issue #35, opened 2022-02-11
Use Dafny CLI instead of lit to verify/test


robin-aws:
(pulling a discussion from the original first big PR out: https://github.com/dafny-lang/libraries/pull/7#discussion_r680374440)

Besides what @samuelgruetter is saying about testing that pre- and post-conditions make sense, I want the CI to ensure the code can be compiled and run and produce the right results. Given that there are lots of `function method`s in this library and not just `function`s, it should be a requirement to ensure that the code is actually compilable and works correctly. 

Here's a really simple example:

```dafny
module SeqTest {
  import Seq
  method {:test} TestApply() {
    var input := [1, 2, 3, 4, 5];
    var output := Seq.apply(i => i * 2, input);
    expect output == [2, 4, 6, 8, 10];
  }
}
```

Building and testing would be a two step process:

```
dafny /compile:0 /spillTargetCode:2 <all Dafny source files>
```

That will still run verification on all the source files and potentially fail. If verification succeeds, you would then execute the resulting C# tests:

```
dotnet test
```

This would require a `.csproj` file similar to the one used in the Dafny test suite: https://github.com/dafny-lang/dafny/blob/master/Test/DafnyTests/DafnyTests.csproj.

The `{:test}` attribute is only handled by the C# compiler for now but we're adding support for the other target languages. Plus, as I mentioned in https://github.com/dafny-lang/dafny/issues/1322#issuecomment-889184653, we should make the process a single simple step by adding something like a `dafny test` command.

Note that this implies dropping the `lit` tool for verifying and testing this library. `lit` is a great tool for testing compilers or CLI tools themselves, but it's a poor match for end-user programs using such tools. We shouldn't expect any of this code to fail to verify or compile, so we don't need the ability to assert the exact expected error/warning behavior. This project can instead serve as an example for best practices in maintaining Dafny code, which should only involve verifying, compiling, and running Dafny code.

Dropping `lit` will mean that we need to change the solution we set up for testing this package in the CI of the `dafny-lang/dafny` repo, but that is definitely doable, and can be handled at the same time that we update the submodule pointer to pick up the new version without `lit` header commands in all the Dafny files.

cpitclaudel:
> Note that this implies dropping the lit tool for verifying and testing this library. 

? Don't our compiler tests use lit currently?  It seems fine for this purpose.

robin-aws:
Lit is great for testing compilers, but this repository is not a compiler. :)

[...]

PR #91, opened 2023-02-17
Updating the original files in libraries to Dafny 4.

davidcok:


robin-aws:
Nit: Mention in the PR body/commit message that you're also adding the autoformatter. I was about to ask if you got the PR title wrong because there are so few changes for Dafny 4.0 compatibility, I wasn't even sure that part was happening. :D

pydiff.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
```
robin-aws, line 1: I don't LOVE adding this file and digging ourselves even deeper into `lit`, but I'm OKAY with it for now. Once we have project files we should definitely tackle #35 though.

robin-aws (changes requested):
Looks great (and I mean that more literally than usual, thank you @MikaelMayer's autoformatter! :)

My only ask is if we're claiming this makes it Dafny 4.0 compatible, we have to back that up in CI. Since we can't just ask `setup-dafny-action` to download 4.0 yet, could we set up lit to support substituting `dafny --functionSyntax:4 --quantifierSyntax:4 --unicode-char ...` as `%dafny` instead? Then we can add a build matrix to the CI to run both with 3.x dafny and with 3.x dafny with those flags.

pydiff.py
davidcok: If we make nightly builds right after 3.12 that have 4.0 defaults, then we can just list a bunch of specific builds to test against. I have tested them locally on my machine, running lit with a 4.0 build.

davidcok:
Using lit made it very easy to setup and update the testing though.

.github/workflows/tests.yml
```diff
@@ -1,5 +1,5 @@
 # This workflow performs static analysis checks.
-name: tests
+name: Legacy verification
```
robin-aws, line 2: Revert since this doesn't make sense without other edits.

.github/workflows/tests.yml
```diff
@@ -17,23 +17,24 @@ jobs:
       - name: Install Dafny
         uses: dafny-lang/setup-dafny-action@v1
         with:
-          dafny-version: "3.9.1"
+          dafny-version: "nightly-2023-02-15-567a5ba"
```
robin-aws, line 20: We shouldn't have the main CI using a nightly build

.github/workflows/tests.yml
```diff
@@ -17,23 +17,24 @@ jobs:
...
-
-      - uses: actions/upload-artifact@v2  # upload test results
-        if: success() || failure()        # run this step even if previous step failed
-        with:
-          name: verification-results
-          path: '**/TestResults/*.trx'
-          if-no-files-found: error
+      - name: Verify Legacy Code
+        run: lit --time-tests -v 
+        ##--param 'dafny_params=/verificationLogger:trx /verificationLogger:csv' .
```
robin-aws, line 30: We shouldn't be disabling parts of the CI, even temporarily.

robin-aws:
Dismissed my earlier Request Changes as I'm okay with not testing in Dafny 4.0 mode for now, since it will be easier to do that once 4.0 is released. I still don't want the CI to be put in a broken state temporarily but I will let other reviewers help address that.

.github/workflows/tests.yml
atomb: It looks like, in particular, the missing `.` at the end is causing the tests to fail at the moment.

atomb:
I think the current test failures may be due to changing the path to be verified from `.` to `src`, since that may have led to the test results being stored in a different directory. I think it makes sense to include the examples in verification, too, which would suggest leaving the target directory as `.`.

davidcok:
Agree that the path should be ., not src. Thanks for spotting that.

atomb:
Hmm. Still failing. If you run the `lit` command locally, do you get any `.trx` files generated?

davidcok:
PR shifted to #97 for a continuing branch

PR #92, opened 2023-02-17
Reorganization of the current library code, under Dafny namespace

davidcok:


src/dafny/DafnyCore.dfy
```diff
@@ -0,0 +1,14 @@
+// RUN: %diff "%s" "%s"
+
+include "./BoundedInts.dfy"
+include "./Results.dfy"
+include "./Collections/Collections.dfy"
+include "./FileIO/FileIO.dfy"
+include "./Math.dfy"
+include "./NonlinearArithmetic"
```
atomb, line 8: Did you mean to create a top-level `.dfy` file for this directory, too? It doesn't seem valid, as-is.

atomb (changes requested):
This looks nice. The one thing I noticed is that the top-level `DafnyCore.dfy` file doesn't seem valid. Maybe a `//RUN: %resolve` directive would be good, to ensure that it at least typechecks?

src/dafny/DafnyCore.dfy
davidcok: I put in the diff commands as No-ops, because lit complains (and fails) if there is no RUN command. But a resolve command would be better. I put the top-level DafnyCore.dfy file, though it has no functionality of its own, in so that a user can include it (as a library) and get the whole library.

I'll change those diffs to resolve where they exist.

davidcok: All of that is fixed up and tests successfully - a few new files.

atomb (approved):
Looks good!

src/dafny/Results.dfy
```diff
@@ -0,0 +1,148 @@
+// RUN: %verify "%s"
+
+/*******************************************************************************
+ *  Copyright by the contributors to the Dafny Project
+ *  SPDX-License-Identifier: MIT 
+ *******************************************************************************/
+
+module {:options "--function-syntax:4"} Dafny.Results {
```
robin-aws, line 8: For the record I still don't think this is an improvement over `Wrappers`, especially not enough to pay the change tax for anyone already using any of this code. But I'll allow this for now just because we're still allowing ourselves to make breaking changes in the future. :)

Scripts/check-examples
```diff
@@ -0,0 +1,367 @@
+#! /usr/bin/env bash
```
robin-aws, line 1: Glad this exists but would prefer not to check it in as dead code without actually hooking it up. Better to cut an issue to follow up on instead.

robin-aws (approved):
I'm not 100% happy with everything in here yet and I have comments on #100 we should address here instead as well, but I recognize that this is a ton of work that we don't want to rot in the meantime. I'm okay with approving this since it only adds on to what's already here, and we're still leaving ourselves the opportunity to make further breaking changes later on.

atomb (approved):

robin-aws (approved):

src/dafny/DafnyCore.dfy
davidcok: Agreed.

.github/workflows/tests.yml
```diff
@@ -3,12 +3,14 @@ name: Dafny Core verification
...
-  schedule:
-    - cron: "7 10 * * *"
   workflow_dispatch:
   pull_request:
     branches: [ master ]
 
+concurrency:
+  group: build-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
```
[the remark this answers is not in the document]
davidcok: Understood. After 4.0 lets talk live and settle outstanding questions.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,69 @@
+# This workflow does static verification of the DafnyCore library
```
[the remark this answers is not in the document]
davidcok: I tried the syntax you suggested (and is in here), but it does not work with bracketed lists.
So I'm planning to do a reusable core to be shared by the two workflows

src/dafny/Results.dfy
davidcok: Note they already have to change to a top-level Dafny module name.

davidcok: An example of how a not-great-but -temporary solution can become permanent

Scripts/check-examples
davidcok: As it is merged, I'll make sure it does not stay dead very long.

PR #97, opened 2023-02-18
Updating the original files in libraries to Dafny 4 (#2)

davidcok:


.github/workflows/tests.yml
```diff
@@ -1,38 +1,66 @@
...
     branches: [ master ]
 
 jobs:
   verification:
-    runs-on: ubuntu-latest
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        # nightly-latest to catch anything that breaks these tests in current development
```
robin-aws, line 18: Definitely happy with the solution of also using a specific nightly build to use the logging! But "nightly-latest" should only be used in the scheduled build, since otherwise the mainline CI can break spontaneously.

Here's how I set it up on the reportgenerator, which I intend to be the template for other projects: https://github.com/dafny-lang/dafny-reportgenerator/pull/17/files You should be able to do something similar, the only difference being that you have to make the matrix variable conditional instead, something like:

```yaml
version: ${{ github.event_name == 'schedule' && ['nightly-latest'] || ['nightly-2023-02-18-ef4f346', '3.11.0'] }}
```

README-TESTING.md
```diff
@@ -0,0 +1,32 @@
...
+of the clone of the repo). You can also manually run individual files or directories,
+but `lit` must be invoked in the `libraries` directory. You can also trigger the workflow
+using `workflow_dispatch`. To do a manual check of formatting, use `dafny format --check <file or folder>`.
+
+The tests themselves are specially formatted lines (those beginning with `// RUN:`) in
+each .dfy file. The tests use dafny's new command-line syntaxi, using the lit
+macros `%verify` and `%run` and the like.
+
+The versions of dafny being run are set in the `tests.yml` file. The versions tested
+ must be recent enough to support the new CLI syntax, so 3.11ff. Also, if
```
robin-aws, line 27: Truncated?

README-TESTING.md
```diff
@@ -0,0 +1,32 @@
+# Testing the libraries
+
+The code in these libraries is tested both with deductive verification, using dafny,
+and with traditional runtime testing of compilations by dafny. 
+The latter is important for those cases (e.g. FileIO) where the implementations
+are custom external code for each target programming language.
+Runtime tests are also a check that dafny compiled the (verified) dafny code correctly.
```
robin-aws, line 7: (minor) I'd be slightly happier reversing the order of these two motivations, just because all Dafny code needs tests because of the second reason, whereas the first is just emphasizing the code most likely to be among the riskiest inside the trusted computing base.

robin-aws (changes requested):
Thanks so much for figuring out how to avoid dropping the testing bar on this one. Just one more tweak to align how Dafny projects should use "nightly-latest"

README-TESTING.md
robin-aws: There's also the reason of making sure we watch out for any code that isn't currently supported for all compilation targets (but we can leave that for now in this PR, since we're not yet claiming anything about this)

.github/workflows/tests.yml
```diff
@@ -1,38 +1,66 @@
...
+        run: |
+          lit --time-tests -v --param 'dafny_params=--log-format trx --log-format csv' .
 
       - name: Generate Report
+        if: always() && steps.withlog.conclusion != 'skipped'
         run: find . -name '*.csv' -print0 | xargs -0 --verbose dafny-reportgenerator summarize-csv-results --max-duration-seconds 10
 
       - uses: actions/upload-artifact@v2  # upload test results
-        if: success() || failure()        # run this step even if previous step failed
+        if: always() && steps.withlog.conclusion != 'skipped'
```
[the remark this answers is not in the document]
davidcok: So that the step runs even if the previous steps fail

.github/workflows/tests.yml
davidcok: OK. But are you comfortable with that tweaking of the test scheduling **after** this PR is merged, so the changes to the script can be properly tested on master-branch files?

README-TESTING.md
davidcok: Odd. Its OK in the Files view. I'll check the rendering again after pushing these changes.

README-TESTING.md
davidcok: Switched the order.

atomb (approved):
I think making the change to `latest-nightly` on another PR (soon) is reasonable, but I'd also like to hear whether @robin-aws has any concerns.

.github/workflows/tests.yml
robin-aws: Sure, it's only going to be a maintenance pain for us if we don't fix it quickly. :)

.github/workflows/tests.yml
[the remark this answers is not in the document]
robin-aws: Then why not JUST always()?

robin-aws (approved):

.github/workflows/tests.yml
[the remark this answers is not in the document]
davidcok: Because if we (intentionally) run with an old version of dafny that does not support logging with new CLI (i.e. 3.11.0) then we don't want this step to run (and fail) if we have not produced any log files.

.github/workflows/tests.yml
davidcok: Put in the next PR. I learn new Github action syntax every day.

davidcok: Actually, this syntax does not work with lists -- tried numerous variations. So for now left out the nightly-latest from the CI build and made a separate nightly build. But this cries out for a reusable workflow that I'll figure out later.

Issue #90, opened 2023-02-16
Request: basic File I/O needs to work for --unicode-char:true


davidcok:

alex-chew:
@robin-aws I think it's possible for the file I/O externs to support both `--unicode-char` options at the same time. The exposed methods use `string` only for the file path parameter and the error message types (the read/written bytes are `seq<bv8>` and not `string`). Does the following approach make sense to you?

- In target languages that support method overloading, like Java and C#, define a method overload for each compiled `char` type (`char` and `Dafny.Rune` in C#, or `char` and `Dafny.CodePoint` in Java)
- In target languages that don't support method overloading but are dynamically typed, like JS and Python, branch on the `path` argument's type and return the error message (if applicable) as the same type
- In Golang, define a union interface of the `_dafny.Char` and `_dafny.CodePoint` wrappers, and (like the JS/Python approach) branch on the argument's actual types (returning an appropriate error message type)

(Python and Golang aren't supported yet, but I want to check that this approach is feasible so we don't have to change our minds when we do add support.)

PR #100, opened 2023-02-23
Documentation for the library

davidcok:
Just the docs.


src/dafny/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+
+An important implementation point about these type definitions is Dafny can determine which native type 
+in the target compiled program best matches the Dafny type. For example, the Dafny type `int16` is
+a signed integer that fits in 16 bits. If a program using this type is compiled to, say Java, then
+variables of this Dafny type will be compiled to Java `short` variables. In some cases there is no
+natural match. For example Java does not have an unsigned 16-bit type while C# and C++ do.
+
+This module defines:
+- unsigned types of 8, 16, 32, 64, 128 bit widths (e.g. `uint32`)
+- signed types of 8, 16, 32, 64, 128 bit widths (e.g. `int16`)
```
robin-aws, line 18:
```suggestion
- unsigned types of 8, 16, 32, 64 bit widths (e.g. `uint32`)
- signed types of 8, 16, 32, 64 bit widths (e.g. `int16`)
```

src/dafny/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+
+Dafny itself generally uses and reasons about mathematical, unbounded integers and natural numbers.
+However compiled programs, for efficiency of computation and storage, will generally use fixed-bit-width
+numbers. So Dafny defines them here, for use when needed.
+
+An important implementation point about these type definitions is Dafny can determine which native type 
+in the target compiled program best matches the Dafny type. For example, the Dafny type `int16` is
+a signed integer that fits in 16 bits. If a program using this type is compiled to, say Java, then
+variables of this Dafny type will be compiled to Java `short` variables. In some cases there is no
+natural match. For example Java does not have an unsigned 16-bit type while C# and C++ do.
```
robin-aws, line 14: YES, thank you! This is worth documenting somewhere with very high visibility in the reference manual/user guide as well.

src/dafny/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+- unsigned types of 8, 16, 32, 64, 128 bit widths (e.g. `uint32`)
+- signed types of 8, 16, 32, 64, 128 bit widths (e.g. `int16`)
+- unsigned types that are subsets of the corresponding signed type (e.g. `nat8` has values from 0 through 127)
+
+The `natN` series of types require some care. A `nat8` for example has non-negative values up through 127,
+that is, just 7-bits worth of values. But it can be directly converted to an `int8` and can be represented by a 
+native signed 8-bit integer type.
+- if you need a general unsigned 8-bit type, with values running up to 256, use `uint8`
+- if you want natural numbers that interact well with signed numbers and do not mind the restriction in range, use `nat8`
+- if the target platform for compilation does not have native unsigned int types, then use nat types because of the smaller range
```
robin-aws, line 26:
```suggestion
- if the target platform for compilation does not have native unsigned int types, then use nat types because of the smaller range.
```

src/dafny/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+
+In addition, the module defines a number of constants that are powers of two (not all of them, just those that are generally useful).
+They are useful in stating the ranges of fixed-bit-width integers. Examples are `TWO_TO_THE_15`, `TWO_TO_THE_32`.
+
+Here are two example methods. In the first, the module `Dafny.BoundedInts` is renamed to `T`, which is then used as the prefix for type and constant names.
```
robin-aws, line 31: nit: in my experience `T` is much more often a type parameter than a module alias. Perhaps `BI` instead?

src/dafny/BoundedInts.md
```diff
@@ -0,0 +1,51 @@
...
+In the second, the module is imported as opened, in which case the type and constant names can be used without qualification.
+<!-- %check-verify -->
```
robin-aws, line 33: This repo doesn't check these inline directives does it? We should either move these examples under `examples` or add that automation.

I think my preference would be the latter, mirroring the way most Dafny projects will define tests/examples, rather than how we do it in the reference manual. But it's worth thinking about supporting something like Python's doctests in Dafny too in the future.

src/dafny/Collections/Collections.md
```diff
@@ -0,0 +1,4 @@
+
+## The `Collections` module {#sec-collections}
+
+_Not yet written_
```
robin-aws, line 4: We can safely delete this and fill in the gap later, right?

src/dafny/DafnyCore.md
```diff
@@ -0,0 +1,51 @@
+
+# Guide to the Dafny Core Library
+
+The Dafny programming language and tool set includes
+libraries that provide functionality in various domains. This document describes 
+the _Dafny Core Library_. Dafny programs are composed of modules; to ensure
```
robin-aws, line 6: Note that this PR isn't intending to relocate code, but all the docs are still sitting under `src/dafny`, which doesn't yet contain any code. You'll have to move them to sit beside their respective Dafny files to make sense.

Given that, I don't actually think you want to talk about the "Dafny Core Library" yet because that only makes sense when referring to `src/dafny` in #92 instead. I think I lean towards just deleting this whole file from this PR and only adding it in #92.

src/dafny/DafnyCore.md
```diff
@@ -0,0 +1,51 @@
...
+verification soundness, Dafny imposes a strict non-circular dependency relationship
+among modules. This Core library contains functionality that may be needed
+by other libraries, but it itself is standalone.
+
+All the supported, delivered libraries will
+- be stable --- a minimum of backwards incompatible changes
+- be verified --- proved to meet specifications and tested with expected use cases,
+with each release of the library and of the Dafny toolset
+- be efficient --- efficient in both proofs and executables that use the libraries
+- be compilable for every supported target language by default, with exceptions being documented and for exceptional reasons
```
robin-aws, line 16: I love these goals but don't want to state them as achieved like this since we don't have mechanisms to back them up yet (especially the latter two), so I'll probably push back on this section even in #92.

src/dafny/FileIO/FileIO.md
```diff
@@ -0,0 +1,122 @@
+
+## The `FileIO` module {#sec-fileio}
+
+_The FileIO methods currently only work with `--unicode-char:false`, which is not the default for Dafny 4._
```
robin-aws, line 4: Good callout, and we should cut an issue to track this.

src/dafny/FileIO/FileIO.md
```diff
@@ -0,0 +1,122 @@
...
+$ dafny run Program.dfy --target:java --input libraries/src/FileIO/FileIO.java
+
+# Javascript
+$ dafny run Program.dfy --target:js --input libraries/src/FileIO/FileIO.js
+```
+
+(If you aren't using `dafny run` to run your program,
+then you should instead integrate the appropriate language-specific implementation file in your build system.)
+
+These sample programs from the examples folder illustrate how to read and write files:
```
robin-aws, line 36: This should be a link instead of making a copy that won't be automatically tested/maintained.

src/dafny/FileIO/FileIO.md
```diff
@@ -0,0 +1,122 @@
+
```
robin-aws, line 1: This will have to be merged with the existing README.md as you move it under the existing FileIO.

src/dafny/Math.md
```diff
@@ -0,0 +1,8 @@
+
+## The `Math` module {#sec-math}
+
+The `Math` module currently contains a few integer-related math functions (e.g. Min, Max, Abs).
+- Functions over integers using multiply, divide and modulo are handled in the
+`NonlinearArith` module, along with lemmas about those functions.
+- Integer-related functions over collections of integers (e.g. minimum of a sequence) are
+available in the `Collections` module.
```
robin-aws, line 8: Good callout - feels weird, but perhaps #92 will put this in a more logical place.

src/dafny/NonlinearArithmetic/NonlinearArithmetic.md
```diff
@@ -0,0 +1,11 @@
+
+## The `NonlinearArith` module {#sec-nonlinear-arith}
+
+Non-linear math is generally undecidable; hence Dafny (and Z3) rely on heuristics to prove such properties.
```
robin-aws, line 4: Can we briefly define non-linear or link somewhere that does?

src/dafny/Relations.md
```diff
@@ -0,0 +1,36 @@
...
+- StronglyConnected : `R(x,y) || R(y,x)`
+- TotalOrdering : Reflexive, AntiSymmetric, Transitive, StronglyConnected (e.g., `<=` on integers)
+- StrictTotalOrdering : Irreflexive, AntiSymmetric, Transitive, Connected (e.g., `<` on integers)
+
+A property of unary functions:
+- Injective : `f(x) == f(y) ==> x == y`
+
+These properties are sometimes required for other functions. For example, 
+if one wants to sort a sequence by some relation `R`, one must establish that `R` is a _Total Ordering_. 
+In fact, that is part of the precondition of the sorting function.
```
robin-aws, line 21: This is slightly tricky since the one sorting function we have (Arrays.BinarySearch) requires a strict total ordering, but #57 will require a total ordering. Ideally we'd clean that up without breaking people, but for now the line above should probably say "Strict Total Ordering" instead.

src/dafny/Relations.md
```diff
@@ -0,0 +1,36 @@
...
+As a simple example, you might define a predicate like this:
+```dafny
+  const IntLT  := ((i: int, j: int) => (i < j))
+```
+
+and then need to proof this lemma to use it in a sorting routine:
+```dafny
+  lemma IntLTisStrictTotalOrder()
+    ensures StrictTotalOrdering(IntLT) {}
+```
```
robin-aws, line 32: Again better to only have under examples - perhaps add the content here as comments there instead?

robin-aws (changes requested):
Wonderful to have so much more documentation on this code, thank you. There are some lingering pieces of #92 in here still - while you're addressing those I'll go review that PR so I have more context in case that helps.

src/dafny/BoundedInts.md
davidcok: Good

src/dafny/BoundedInts.md
davidcok: OK- changed

src/dafny/BoundedInts.md
davidcok: No it doesn't, but that is a one-line change in doc-tests.yml (in the dafny repo testing) to turn on such checking once this PR is merged.

src/dafny/Collections/Collections.md
davidcok: We could. But this is clearer to people looking that the material is coming, rather than we just forgot about it.

src/dafny/DafnyCore.md
davidcok: The intention was to reorganize the code and document the reorganization, rather than write documentation only to have to change it. Also, this file is the starting point for the other material. Without it, users won't have a way to get to it to comment.

And yes, the files are in these subdirectories so they sit next to their code (as you requested before). It does not hurt to have them there now with empty direcotries.

src/dafny/DafnyCore.md
davidcok: I'll state them as the goals we are aiming for -- which is what I intended.

src/dafny/FileIO/FileIO.md
davidcok: There already is such an issue: #90

src/dafny/FileIO/FileIO.md
davidcok: Both will be tested. But post-4.0 when I do some examples/tests cleanup and enhancement, I'll clean this up as well.

src/dafny/FileIO/FileIO.md
davidcok: The existing README.md is going away as it has been subsumed by FileIO.md. Just doing that in two steps.

src/dafny/Math.md
davidcok: It made sense to have the properties like minimum of a sequence in the Seq module -- the various lemmas that were needed to prove the functionality were there anyway.

src/dafny/NonlinearArithmetic/NonlinearArithmetic.md
davidcok: Brief sentence included here.

src/dafny/Relations.md
davidcok: Revised

src/dafny/Relations.md
davidcok: Other reviewers (of other documentation) have predominantly preferred having examples directly present, both for reading and reviewing. I'll add the annotation to be sure these are checked in place.

src/dafny/Results.md
```diff
@@ -0,0 +1,178 @@
...
+An alternative is to write
+```dafny
+  var convert := match o case Some(v) => Success(v)
+                                     case None => Failure("not found") );
+  var index: int :- convert(Find(s, 0, value));
+```
+
+### Combining different FC-types in methods
+
+_This is an experimental, tentatively planned, not yet implemented feature._
```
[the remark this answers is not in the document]
davidcok: Correct - that line should have gone out when the workaround description came in.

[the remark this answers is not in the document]
davidcok: Deleted.

src/dafny/Results.md
```diff
@@ -0,0 +1,178 @@
+
+## The `Results` module {#sec-results}
```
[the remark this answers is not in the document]
davidcok: OK. Though this was the name Ryan suggested.

src/dafny/BoundedInts.md
robin-aws: That's the wrong place to test changes to this repo's contents though. Would it be easy to set up the same checking here instead?

robin-aws: (and fine to add that later if so)

src/dafny/Collections/Collections.md
robin-aws: Ah fair enough, I thought this was meant as a TODO.

src/dafny/Unicode/Unicode.md
```diff
@@ -0,0 +1,4 @@
+
```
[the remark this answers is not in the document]
robin-aws: As above never mind :)

src/dafny/Relations.md
robin-aws: Sure, as above I could see steering this repo towards having all inline examples instead of the current setup, and IF it is possible to set up the mechanism to test them I'm happy with that.

davidcok: It is possible -- just like it is already in the other repo. But it has to be checked in and next to the new code.
Then I'll be able to create a yml to do the checking regularly or on demand.

src/dafny/BoundedInts.md
davidcok: Yes - In the process of adapting the code to do that. We'll get it working in a subsequent PR.

PR #104, opened 2023-03-03
Adding yml to test examples embedded in .md files

davidcok:
- Addint code to tests examples; also reverting the name of Wrappers back to Wrappers


Scripts/check-examples
```diff
@@ -67,6 +67,9 @@
 ##   check
```
robin-aws, line 67: This script snuck in within a previous PR (#92) and I didn't have the chance to look it over much at the time.

Despite my comments on #35, I have no concerns about using something like this script to test embedded examples, and as I commented elsewhere I don't actually think you're using lit here anyway, just some lit-like syntax.

My only immediate concern is having to maintain a copy of this script in every repo that wants to test embedded examples. Why don't we move it under https://github.com/dafny-lang/dafny/tree/master/Scripts instead, marked as experimental? We can just have this CI git clone dafny so it can use that copy. 

(In fact you could make this work with its current location, but I'd like to be more obvious that consumers outside of the core dafny repo are referencing the script)

src/Collections/Sequences/Seq.dfy
```diff
@@ -664,7 +664,7 @@ module {:options "-functionSyntax:4"} Seq {
   /* Filtering a sequence is distributive over concatenation. That is, concatenating two sequences 
      and then using "Filter" is the same as using "Filter" on each sequence separately, and then 
      concatenating the two resulting sequences. */
-  lemma {:opaque} LemmaFilterDistributesOverConcat<T>(f: (T ~> bool), xs: seq<T>, ys: seq<T>)
+  lemma {:opaque} LemmaFilterDistributesOverConcat<T(!new)>(f: (T ~> bool), xs: seq<T>, ys: seq<T>)
```
robin-aws, line 667: Why is this changing in this PR?

robin-aws (changes requested):

src/Collections/Sequences/Seq.dfy
davidcok: Because running the tests elicited a complaint about needing the !new -- I'm guessing from the work on allocated:4.

Scripts/check-examples
davidcok: The script is customized - slightly different than the one in dafny-lang/dafny. I'd like to get these tests working & merged and in a separate step refactor the script to be useful in multiple locations.

Created issue #3688 to address this.

robin-aws: Ah that's fine then, will continue discussion on https://github.com/dafny-lang/dafny/issues/3688

src/Collections/Sequences/Seq.dfy
robin-aws: I think you must have run with 3.13 or later at some point then, since that's where that change happened, but this repo's CI was already running with 3.12 so it wouldn't have broken earlier.

Is it possible to keep this verifying without adding the type characteristic, since this might break consumers?

davidcok: That would take verifying different releases of the library with different releases of dafny. We probably ought to make such infrastructure, but we don't have it yet. That would be joined with carefully documenting which releases of the library go with which dafny releases. At the moment we don't even officially release the library (i.e. have releasae numbers).

But yet, I was running tests using a current (master) build of dafny, encountered the !neww problem and fixed it in anticipation of adding 3.13.1 and 4.0.0 into the tests mix.I expected that it would be considered unsound to use that lemma for a type that was not !new even if there is no check to tell you so.

robin-aws: Ah thanks, I was confused because I tried this with those versions and didn't see an error on that lemma. There WAS an issue with Filter's older `reads` clause but that's been fixed.

There's no quantification over `T` in this lemma so I think you can revert this. If you see an actual unsoundness we need to fix upstream Dafny instead!

davidcok: Actually the problem was that the version in the original library was changed when allocated:4 went in, but not the copy in the reorg. I changed the reorg copy to match.

src/dafny/DafnyCore.md
```diff
@@ -23,7 +23,7 @@ in the github repository [here](https://github.com/dafny-lang/libraries/tree/mas
 ## Content of the Core Library
 
 The sections below describe how to use each of the principal modules of the library:
-- [Dafny.Wrappers](Results) -- containers that enable encapsulaton of success and failure results in a single datatype
+- [Dafny.Wrappers](Wrappers) -- containers that encapsulate success and failure results in a single datatype
```
robin-aws, line 26: This description doesn't match the content, I'm taking a stab at a different one...

robin-aws (changes requested):
Partial review for now...

src/dafny/DafnyCore.md
robin-aws:
```suggestion
- [Dafny.Wrappers](Wrappers) -- simple datatypes to support common patterns, such as optional values or the result of operations that can fail.
```

robin-aws (changes requested):
Thanks for accepting my one-liner describing the contents of `Wrappers`. There's another similar phrase at the top of `src/dafny/Wrappers.md` we should replace as well (I couldn't comment directly on that line since it wasn't edited in this PR)

robin-aws (dismissed):

.github/workflows/nightly.yml
```diff
@@ -15,12 +15,11 @@ jobs:
...
         # nightly-latest to catch anything that breaks these tests in current development
-        # 2/18/2023 version is the first that supports logging
+        # 2/18/2023 version is the first that supports logging, but it is not supported by setup-dafny-action 1.6.1
         # 3.11.0 supports new CLI but does not support logging
         # setup-dafny-action does not yet support 3.13.1 or recent nightly-lates
-t
-        ## version: [ nightly-latest, nightly-2023-02-18-ef4f346, 3.11.0, 3.12.0, 3.13.1 ]
-        version: [ nightly-2023-02-18-ef4f346, 3.11.0, 3.12.0 ]
+
+        version: [ nightly-latest, nightly-2023-02-18-ef4f346, 3.11.0, 3.12.0i, 3.13.1, 4.0.0 ]
```
atomb, line 22:
```suggestion
        version: [ nightly-latest, nightly-2023-02-18-ef4f346, 3.11.0, 3.12.0, 3.13.1, 4.0.0 ]
```
Looks like a typo?

.github/workflows/tests.yml
```diff
@@ -14,11 +14,10 @@ jobs:
     strategy:
       matrix:
         # nightly-latest to catch anything that breaks these tests in current development
-        # 2/18/2023 version is the first that supports logging
+        # 2/18/2023 version is the first that supports logging, but it is not supportied by dafny-setup-action 1.6.1
```
atomb, line 17:
```suggestion
        # 2/18/2023 version is the first that supports logging, but it is not supported by dafny-setup-action 1.6.1
```
Typo, though it's not really important.

atomb (changes requested):
Looks good, with the exception of a few tiny typos (one of which seems significant).

atomb (approved):

PR #105, opened 2023-03-06
Equivalence relation

stefan-aws:


examples/Relations.dfy
```diff
@@ -18,4 +18,15 @@ module RelationsExamples {
   lemma BuiltInIntLEIsTotalOrdering()
     ensures TotalOrdering((x: int, y: int) => x <= y)
   {}
+
+  function Modulo(n: nat): (R: (int, int) -> bool)
+    requires n > 1
```
davidcok, line 23: Why not allow 1?

davidcok (changes requested):
Be sure to add some examples that serve as tests.

examples/Relations.dfy
stefan-aws: Fixed

jtristan (approved):

davidcok (approved):
Since John has already approved, I'll add an approval.
But please add this change in src/dafny also -- and make any future changes there.

PR #110, opened 2023-03-10
Extract element from singleton set

stefan-aws:


davidcok (changes requested):
Should have  an example that uses it -- in particular one that shows why this method should be used over simply `x :| x in s`

src/Collections/Sets/Sets.dfy
```diff
@@ -76,6 +76,36 @@ module {:options "-functionSyntax:4"} Sets {
...
+    }
+    if IsSingleton(s) {
+      var x :| x in s;
+      assert s == {x};
+      assert |s| == 1;
+    }
+  }
+
+  /* Extracts the unique element from a singleton set. */
+  method ExtractFromSingleton<T>(s: set<T>) returns (x: T)
```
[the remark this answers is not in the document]
stefan-aws: Changed it to `function` and added context

src/Collections/Sets/Sets.dfy
atomb, line 1: What's the long-term plan for these parallel trees? It seems like they'll be hard to maintain.

atomb (approved):

src/Collections/Sets/Sets.dfy
stefan-aws: My understanding is that as we move forward, new code to the library should only be added to the `dafny` subfolder.

PR #113, opened 2023-03-23
Unique_maximum

stefan-aws:


src/Collections/Sets/Sets.dfy
```diff
@@ -193,4 +195,36 @@ module {:options "-functionSyntax:4"} Sets {
     LemmaSubsetSize(x, range);
   }
 
+  /* Any totally ordered set contains a unique minimal element. */
```
davidcok, line 198: Start starting these comments with /** so that they will be docstsrings.

davidcok (dismissed):

src/dafny/Relations.dfy
```diff
@@ -52,4 +52,13 @@ module {:options "-functionSyntax:4"} Dafny.Relations {
     && Connected(R)
   }
 
+  /** An element in an ordered set is called minimal, if it is less than every other element of the set. */
+  ghost predicate IsMinimum<T>(R: (T, T) -> bool, m: T, s: set<T>) {
```
keyboardDrummer, line 56: Could you give a more descriptive name to the relation, like `lessThanOrEquals` ?

src/dafny/Relations.dfy
```diff
@@ -52,4 +52,13 @@ module {:options "-functionSyntax:4"} Dafny.Relations {
...
+    m in s && forall y | y in s :: R(m, y)
```
keyboardDrummer, line 57: The variable name `m` and `y` aren't very descriptive. I would use `element` and `other`, respectively.

src/dafny/Collections/Sets.dfy
```diff
@@ -193,4 +193,35 @@ module {:options "-functionSyntax:4"} Dafny.Collections.Sets {
     LemmaSubsetSize(x, range);
   }
 
+  /** Any totally ordered set contains a unique minimal element. */
```
keyboardDrummer, line 196: The documentation mentions `minimal` elements, but the specification only says something about minimum elements.

src/dafny/Collections/Sets.dfy
```diff
@@ -193,4 +193,35 @@ module {:options "-functionSyntax:4"} Dafny.Collections.Sets {
...
+  lemma LemmaUniqueMinimum<T(!new)>(R: (T, T) -> bool, s: set<T>) returns (m: T)
```
keyboardDrummer, line 197: Shouldn't you have a lemma that proves that for any set, even partially ordered ones, there is at most one minimum, and a separate lemma that states that for a totally ordered set, there is at least one minimum?

src/Collections/Sets/Sets.dfy
```diff
@@ -193,4 +195,36 @@ module {:options "-functionSyntax:4"} Sets {
     LemmaSubsetSize(x, range);
   }
 
+  /** Any totally ordered set contains a unique minimal element. */
+  lemma LemmaUniqueMinimum<T(!new)>(R: (T, T) -> bool, s: set<T>) returns (m: T)
```
keyboardDrummer, line 199: Where does the duplication between `src/Collections/Sets/Sets.dfy` and `src/dafny/Collections/Sets.dfy` come from?

src/dafny/Collections/Sets.dfy
stefan-aws: In my opinion, it is fine to capture uniqueness in the post-condition of `LemmaUniqueMinimum`, as it is done right now: `ensures IsMinimum(R, m, s) && (forall n: T | IsMinimum(R, n, s) :: m == n)`.

src/dafny/Relations.dfy
stefan-aws: Note that all predicates in `Relations.dfy` are currently using the name `R`. In my opinion, using `lessThanOrEquals` instead is misleading (and too long), as we are talking about _generic_ relations, not the particular case where the relation is given by the familiar `<=` relation (on e.g. the naturals), which satisfies specific properties.

src/Collections/Sets/Sets.dfy
stefan-aws: https://github.com/dafny-lang/libraries/pull/92

src/dafny/Relations.dfy
```diff
@@ -52,4 +52,13 @@ module {:options "-functionSyntax:4"} Dafny.Relations {
     && Connected(R)
   }
 
+  /** An element in an ordered set is called minimal, if it is less than every other element of the set. */
```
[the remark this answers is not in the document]
stefan-aws: I had a look at the definitions again. As far as I understand, an element `x in (S,<=)` is:
- the _minimal_ element of `S`, if for all `y in S`, we have `y <= x` implies `x == y`;
- the _minimum_ element of `S`, if for all `y in S`, we have `x <= y`.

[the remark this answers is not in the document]
keyboardDrummer: I agree with that, but I don't see how it responds to my comment.

In any case, if the name mentions `Minimum` than the documentation should refer to the same concept.

Also, the predicate `IsMinimum` doesn't have a total ordering as a precondition on the relation `R`, so minimum and minimal are not the same here. The definition and name correspond to the minimum element, so the documentation should as well.

src/dafny/Relations.dfy
keyboardDrummer: Sounds good :)

src/dafny/Collections/Sets.dfy
keyboardDrummer: I think it's fine too. But just to be clear, proving that for general relations, without a total ordering constraint, there is at most one minimum, gives a more generally applicable lemma, right?

src/dafny/Relations.dfy
keyboardDrummer: Just some extra context:

My hope was that using the name `lessThanOrEquals` would _only_ indicate that IsMinimum is true if the relation always holds when `m` is on the _left_ of the relation, not anything else like whether the relation is totally or partially ordered. Without that name, how is a user supposed to know when to use `IsMinimum` vs `IsMaximum`?

If I look at the Wiki definition of `minimum` for a generic relation, it also uses the terms "greater than or equal to", without implying anything about partially vs totally ordered.

src/dafny/Collections/Sets.dfy
stefan-aws: True, to prove uniqueness, anti-symmetry should be sufficient.

src/dafny/Relations.dfy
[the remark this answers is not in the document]
stefan-aws: Apologies, I got distracted before I was able to follow up on my own comment above. In short, I agree with you; I'll clarify!

stefan-aws:
Thanks for the hints, I added additionally material and now distinguish between:
- minimal 
- least (minimum)

and

- maximal
- greatest (maximum)

src/dafny/Collections/Sets.dfy
stefan-aws: I added separate uniqueness proofs.

src/dafny/Relations.dfy
```diff
@@ -56,9 +56,42 @@ module {:options "-functionSyntax:4"} Dafny.Relations {
     && Connected(R)
   }
 
+  ghost predicate PreOrdering<T(!new)>(R: (T, T) -> bool) {
```
keyboardDrummer, line 59: What's a preordering and why is this required for some of the lemma's?

keyboardDrummer (approved):

atomb (approved):
This looks very nice, even more so after the latest improvements.

PR #115, opened 2023-03-23
Add Equivalence Relation Definitions to Dafny Subfolder

stefanzetzsche:
Follow up to https://github.com/dafny-lang/libraries/pull/105

### DIFF
--- a/examples/BinaryOperations.dfy
+++ b/examples/BinaryOperations.dfy
@@ -1,0 +1,182 @@
+// RUN: %verify "%s"
+
+/*******************************************************************************
+ *  Copyright by the contributors to the Dafny Project
+ *  SPDX-License-Identifier: MIT
+ *******************************************************************************/
+
+include "../src/dafny/BinaryOperations.dfy"
+
+module {:options "-functionSyntax:4"} IntegersExample {
+
+  import opened BinaryOperations
+
+  ghost function add(x: int, y: int): int {
+    x + y
+  }
+
+  ghost function minus(x: int): int {
+    -x
+  }
+
+  ghost function mult(x: int, y: int): int {
+    x * y
+  }
+
+  lemma IntegersAreAssociative()
+    ensures IsAssociative(add)
+    ensures IsAssociative(mult)
+  {}
+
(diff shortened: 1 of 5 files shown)


keyboardDrummer (approved):

src/dafny/Relations.dfy
```diff
@@ -26,6 +26,10 @@ module {:options "-functionSyntax:4"} Dafny.Relations {
     forall x, y :: R(x, y) && R(y, x) ==> x == y
   }
 
+  ghost predicate Symmetric<T(!new)>(R: (T, T) -> bool) {
+    forall x, y :: R(x, y) <==> R(y, x)
```
keyboardDrummer, line 30: Just a question for my own education, what's the difference between using `<==>` and `==` here?

src/dafny/Relations.dfy
```diff
@@ -26,6 +26,10 @@ module {:options "-functionSyntax:4"} Dafny.Relations {
     forall x, y :: R(x, y) && R(y, x) ==> x == y
   }
 
+  ghost predicate Symmetric<T(!new)>(R: (T, T) -> bool) {
```
keyboardDrummer, line 29: What's the reason a predicate like this shouldn't be used for heap based types? (`!new`)

src/dafny/Relations.dfy
stefan-aws: According to Section 5.2.1.1. in https://dafny.org/dafny/DafnyRef/DafnyRef, _the expressions `A <==> B` and `A == B` give the same value, but note that `<==>` is associative whereas `==` is chaining and they have different precedence_.

src/dafny/Relations.dfy
stefan-aws: The reason is quantification: removing `!new` leads to the error: _a forall expression involved in a predicate definition is not allowed to depend on the set of allocated references, but values of 'x' (of type 'T') may contain references (perhaps declare its type as 'T(!new)') (see documentation for 'older' parameters)_.

Issue #111, opened 2023-03-13
Request: connecting Unicode logic to the string type


robin-aws:
`src/Unicode` has a fairly pure implementation of UTF-16 and UTF-8, but it uses bit vector types and doesn't use `char` or `string` at all. We definitely need a standard library that does (and the incoming JSON PR does itself, since it needs to encode a `string` in UTF-8), although this means we'll need at least some Dafny code that is duplicated to account for the difference in behavior from `--unicode-char`.

robin-aws:
Addressed by #51 - see https://github.com/dafny-lang/libraries/blob/master/src/Unicode/UnicodeStrings.dfy

PR #118, opened 2023-04-12
Generalize JSON code to work with either `--unicode-char` mode

robin-aws:
Replaced the `src/JSON/Utils/Unicode.dfy` code with references to the `Unicode` library, not only because the latter has more correctness proofs but also because it more safely fails to translate invalid data rather than using replacement characters. This introduces more cases of needing `Result` wrappers, but most of the library needed them already anyway so it does not impact usability. I've also likely introduced some minor performance regressions compared to the current state of `json-merge` by using the less-optimized `Unicode` library, but they should be worth it so this code can be merged, since performance can always be improved later.

Adds a generic `UnicodeStrings` module to help with this, which should be helpful for other codebases that need to support both modes. This also addresses #111.

Open question: which of the two options in `UnicodeStrings.dfy` should we use in this repository in general?


davidcok (changes requested):
Please put these changes under src/dafny -- that is the copy of the library we are trying to move ahead with.

robin-aws:
> Please put these changes under src/dafny -- that is the copy of the library we are trying to move ahead with.

More relevant to address in #51 - this one just builds on that one

PR #123, opened 2023-05-12
fix: Avoid stack overflow in Unicode with function-by-method

MikaelMayer:


src/Unicode/UnicodeEncodingForm.dfy
```diff
@@ -109,27 +109,62 @@ abstract module {:options "-functionSyntax:4"} UnicodeEncodingForm {
...
+      PartitionCodeUnitSequenceCheckedTailRec(s[|prefix|..], resultPrefix + [prefix], sPrefix + prefix)
+  }
+
   /**
     * Returns the unique partition of the given code unit sequence into minimal well-formed code unit subsequences,
     * or None if no such partition exists.
     */
   function PartitionCodeUnitSequenceChecked(s: CodeUnitSeq): (maybeParts: Option<seq<MinimalWellFormedCodeUnitSeq>>)
-    ensures maybeParts.Some? ==> Seq.Flatten(maybeParts.Extract()) == s
+    ensures maybeParts.Some? ==> && Seq.Flatten(maybeParts.Extract()) == s
```
atomb, line 137: Was this a temporary change that's no longer needed?

alex-chew, line 137: Why is this important?

src/Unicode/UnicodeEncodingForm.dfy
```diff
@@ -109,27 +109,62 @@ abstract module {:options "-functionSyntax:4"} UnicodeEncodingForm {
     }
   }
 
+  /**
+    * Returns the unique partition of the given code unit sequence into minimal well-formed code unit subsequences,
+    * or None if no such partition exists.
+    */
+  function {:tailrecursion} PartitionCodeUnitSequenceCheckedTailRec(s: CodeUnitSeq, resultPrefix: seq<MinimalWellFormedCodeUnitSeq> := [], ghost sPrefix: CodeUnitSeq := []): (maybeParts: Option<seq<MinimalWellFormedCodeUnitSeq>>)
```
atomb, line 116: This function is never used. Do you intend that clients may call it, or is it a leftover from a previous approach?

alex-chew (approved):
LGTM, just one nitpick comment

src/Unicode/UnicodeEncodingForm.dfy
```diff
@@ -109,27 +109,62 @@ abstract module {:options "-functionSyntax:4"} UnicodeEncodingForm {
...
-      else
-        var prefix := maybePrefix.Extract();
-        // Recursing on subsequences leads to quadratic running time in most/all Dafny runtimes as of this writing.
-        // This definition (and others in the Unicode modules) emphasizes clarify and correctness,
-        // but should be supplemented with a by-method for improved performance,
-        // so long as Dafny runtimes' lack optimizations for subsequence recursion.
-        var restParts := PartitionCodeUnitSequenceChecked(s[|prefix|..]);
-        if restParts.Some? then Some([prefix] + restParts.Extract())
-        else None
+      var prefix :- SplitPrefixMinimalWellFormedCodeUnitSubsequence(s);
```
atomb, line 142: Nice 🐘 use.

atomb (changes requested):
This is a very nice improvement. I saw two things that I think may have been artifacts of previous attempts that could potentially be cleaned up, but I think it's otherwise ready to go.

src/Unicode/UnicodeEncodingForm.dfy
robin-aws: This was my first attempt that broke callers (since they relied on the non-opaque function definition), I'm removing it.

atomb:
It looks like there's a CI formatting failure and a missing `RUN` line on your new test.

src/Unicode/UnicodeEncodingForm.dfy
robin-aws: And I just realized the next two lines can use an 🐘 instead too...

robin-aws:
> It looks like there's a CI formatting failure and a missing `RUN` line on your new test.

Yup, that's what I was working on fixing. :)

robin-aws (dismissed):

robin-aws (approved):

atomb (approved):
🎉

src/Unicode/UnicodeEncodingForm.dfy
```diff
@@ -119,17 +119,35 @@ abstract module {:options "-functionSyntax:4"} UnicodeEncodingForm {
...
+      // The best solution would be to ensure all Dafny runtimes implement subsequences as an O(1)
+      // operation, so this implementation would become linear.
+      var restParts :- PartitionCodeUnitSequenceChecked(s[|prefix|..]);
+      Some([prefix] + restParts)
+  } by method {
+    if s == [] {
+      return Some([]);
+    }
+    var result: seq<MinimalWellFormedCodeUnitSeq> := [];
+    var rest := s;
```
cpitclaudel, line 136: Would keeping an offset here instead of a full suffix make this function faster?

MikaelMayer: Only if we refactored `SplitPrefixMinimalWellFormedCodeUnitSubsequence` was able to take an index and the full sequence, otherwise the call below still has to take the full suffix.
But otherwise that's  a good idea in general.

PR #124, opened 2023-05-17
Changing forall ensures to assert-by to avoid deprecation warning

davidcok:
Fixes issue 3559 in dafny-lang/dafny


davidcok:
Problem: Some modules do not prove under 3.13.1

PR #127, opened 2023-05-17
Fix JSON proof instability

stefan-aws:
`src/JSON/ZeroCopy/Deserializer.dfy` and `src/JSON/ZeroCopy/Serializer.dfy` verified in 4.0, but not in 4.1. This PR fixes the issue.


davidcok (approved):
I presume it tests OK - so fine.

PR #131, opened 2023-07-10
fix: Partially fix nightly build for 4.2

robin-aws:
Only fixes the fact that the top-level workflow looks green even though the job fails (https://github.com/dafny-lang/libraries/actions/runs/5506948866) and adds `--compile-suffix` when necessary.

### DIFF
--- a/.github/workflows/check-examples-in-docs.yml
+++ b/.github/workflows/check-examples-in-docs.yml
@@ -11,7 +11,6 @@ concurrency:
 
 jobs:
   check-examples-in-docs:
-    continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/reusable-tests.yml
+++ b/.github/workflows/reusable-tests.yml
@@ -10,7 +10,6 @@ on:
 
 jobs:
   reusable_verification:
-    continue-on-error: true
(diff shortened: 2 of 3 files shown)


atomb (approved):
I'm on board with merging this even though the 4.1 tests fail, to make sure that our next PR has to be one to fix those tests. However, do the settings on this repository allow us to merge a PR with failing tests?

robin-aws:
> I'm on board with merging this even though the 4.1 tests fail, to make sure that our next PR has to be one to fix those tests. However, do the settings on this repository allow us to merge a PR with failing tests?

When I cut this PR I thought I was just making progress on working against the latest nightly, but now I see that 4.1 has never worked. :(

I'll merge this by temporarily changing the branch protection rules, just because I think turning the nightly build red as it should be is important for visibility as well.

MikaelMayer (approved):
Makes sense, thanks!

atomb (approved):
Let's see if we can get things working with 4.1 soon, but this is still better than it was. At least every PR gets the notification that it's still broken.

robin-aws:
I've re-added 4.1 as a required check, so the CI is effectively broken on this repo now as intended. :)

PR #133, opened 2023-07-14
Fix Json verification

stefan-aws:


src/JSON/Utils/Cursors.dfy
```diff
@@ -142,6 +142,15 @@ module {:options "-functionSyntax:4"} JSON.Utils.Cursors {
     function Split() : (sp: Split<View>) requires Valid?
       ensures sp.SplitFrom?(this, (v: View) => v.Bytes())
       ensures beg != point ==> sp.StrictlySplitFrom?(this, (v: View) => v.Bytes())
+      ensures !BOF? ==> (sp.StrictlySplitFrom?(this, (v: View) => v.Bytes()) && sp.cs.StrictSuffixOf?(this))
```
atomb, line 145: Given how much you're adding to the postconditions, would it be feasible to make `Split` opaque, as well?

src/JSON/Utils/Cursors.dfy
```diff
@@ -265,6 +274,11 @@ module {:options "-functionSyntax:4"} JSON.Utils.Cursors {
       decreases SuffixLength()
       ensures ps.AdvancedFrom?(this)
       ensures !EOF? ==> ps.StrictlyAdvancedFrom?(this)
+      ensures ps.s == this.s
```
atomb, line 277: Similar comment as for `Split`.

src/JSON/ZeroCopy/Deserializer.dfy
```diff
@@ -70,16 +79,24 @@ module {:options "-functionSyntax:4"} JSON.ZeroCopy.Deserializer {
...
       var SP(before, cs) := WS(cs);
       var SP(val, cs) := cs.SkipByte().Split();
       var SP(after, cs) := WS(cs);
       SP(Grammar.Structural(before, val, after), cs)
     }
 
-    type ValueParser = sp: SubParser<Value> |
-        forall t :: sp.spec(t) == Spec.Value(t)
-      witness *
+    ghost predicate ValueParserValid(sp: SubParser<Value>) {
```
atomb, line 95: This seems like a good candidate to make opaque, as well, given the quantifier, though opacity in subtypes can be tricky.

src/JSON/Utils/Cursors.dfy
stefan-aws: Removed most ensures clauses.

src/JSON/Utils/Cursors.dfy
stefan-aws: Removed most ensures clauses.

src/JSON/ZeroCopy/Deserializer.dfy
stefan-aws: Unfortunately turning the predicate opaque caused a couple of issues. For now, I propose we keep it as it is.

src/JSON/ConcreteSyntax.SpecProperties.dfy
```diff
@@ -12,23 +12,20 @@ module {:options "-functionSyntax:4"} JSON.ConcreteSyntax.SpecProperties
...
-      | && (forall d | d < bracketed :: pd0.requires(d))
-      && (forall d | d < bracketed :: pd1.requires(d))
-      && (forall d | d < bracketed :: pd0(d) == pd1(d))
-    {
-      calc {
-        Spec.Bracketed(bracketed, pd0);
-        { ConcatBytes_Morphism(bracketed.data, pd0, pd1); }
-        Spec.Bracketed(bracketed, pd1);
-      }
+    calc {
```
atomb, line 25: Nice simplification!

atomb (dismissed):
Looks nice to me!

src/JSON/ZeroCopy/Deserializer.dfy
```diff
@@ -259,9 +278,19 @@ module {:options "-functionSyntax:4"} JSON.ZeroCopy.Deserializer {
...
+        && ((!cs.BOF? || !cs.EOF?) && (s0 == SEPARATOR as opt_byte) ==> (var sp: Split<Structural<jcomma>> := sp; sp.cs.StrictSuffixOf?(cs)))
+        && ((s0 == SEPARATOR as opt_byte) ==> var sp: Split<Structural<jcomma>> := sp; sp.SplitFrom?(cs, st => Spec.Structural(st, SpecView)))
+        && ((!cs.BOF? || !cs.EOF?) && (s0 == CLOSE as opt_byte) ==> (var sp: Split<Structural<jclose>> := sp; sp.cs.StrictSuffixOf?(cs)))
+        && ((s0 == CLOSE as opt_byte) ==> var sp: Split<Structural<jclose>> := sp; sp.SplitFrom?(cs, st => Spec.Structural(st, SpecView)))
+    {}
+
     // The implementation and proof of this function is more painful than
     // expected due to the tail recursion.
-    function {:opaque} {:tailrecursion} Elements(
+    function {:timeLimit 30} {:vcs_split_on_every_assert} {:opaque} {:tailrecursion} Elements(
```
robin-aws, line 293:
```suggestion
    function {:vcs_split_on_every_assert} {:opaque} {:tailrecursion} Elements(
```

(And likewise below)

I assume this was here for debugging earlier. We should make sure we only include very generous wall-clock timeouts to maximize build reproducibility, and I don't see a huge value in bounding this more tightly (we could set an `:rlimit` instead if that's the intention)

src/JSON/ZeroCopy/Deserializer.dfy
```diff
@@ -276,49 +305,110 @@ module {:options "-functionSyntax:4"} JSON.ZeroCopy.Deserializer {
...
-          assert elem.StrictlySplitFrom?(elems.cs, ElementSpec);
-        }
-        var elems' := AppendLast(open.cs, json, elems, elem, sep);
-        assert elems'.SplitFrom?(open.cs, SuffixedElementsSpec) by {
-          assert elems'.StrictlySplitFrom?(open.cs, SuffixedElementsSpec);
-        }
-        var bracketed := BracketedFromParts(cs0, open, elems', sep);
-        assert bracketed.StrictlySplitFrom?(cs0, BracketedSpec);
-        Success(bracketed)
+      if elem.cs.EOF? then
```
robin-aws, line 308: Just for my understanding, was the key improvement re-ordering the cases or was this a deeper optimization of some kind?

src/JSON/ZeroCopy/Deserializer.dfy
```diff
@@ -426,19 +517,82 @@ module {:options "-functionSyntax:4"} JSON.ZeroCopy.Deserializer {
           assert SP(arr, cs').StrictlySplitFrom?(cs, Spec.Array);
           Spec.UnfoldValueArray(v);
         }
+        assert sp.StrictlySplitFrom?(cs, Spec.Value);
         Success(sp)
       else if c == '\"' as opt_byte then
-        var SP(str, cs) :- Strings.String(cs);
-        Success(SP(Grammar.String(str), cs))
+        var SP(str, cs') :- Strings.String(cs);
```
robin-aws, line 523: Worth a comment about why all the expansion/duplication is necessary here?

robin-aws (changes requested):
Very nice improvements! Just the one blocking issue with `:timeLimit` from me

src/JSON/ZeroCopy/Deserializer.dfy
stefan-aws: The case split is new I believe. As you can see from the Lemma `AboutTryStructural` I added, it is important to know that `!elemen.cs.EOF?`.

src/JSON/ZeroCopy/Deserializer.dfy
stefan-aws: The cases have been there previously. I added the calc's such that post-condition of `Value` goes through. Each of the proofs requires prior knowledge about the shape of `c`, to use either `Grammar.String` or `Grammar.Bool`, and either `var f := _ => TRUE;` or `var f := _ => FALSE;` etc. I think there is not much duplication we can remove here, unfortunately.

src/JSON/ZeroCopy/Deserializer.dfy
stefan-aws: I moved the Lemma `AboutTryStructural` to the actual `assertion by` clauses it is needed for.

src/JSON/ZeroCopy/Deserializer.dfy
stefan-aws: You are right, it's been used for debugging. I'll remove it everywhere.

atomb (approved):
Looks good!

robin-aws (approved):
🚀

PR #142, opened 2023-09-19
fix: missing type characteristic for a functions that accesses the Va…

jtristan:
…lues of a map


jtristan:
Function is ghost, so the problem is in the resolver

Issue #145, opened 2023-09-29
JSON regression test fails


markrtuttle:
I'm unable to run the JSON regression test in Tests.dfy with dafny 4.2.0 from Homebrew on MacOS.  

Tests.dfy begins with
```
// RUN: %run "%s" --unicode-char:false --input ../Unicode/UnicodeStringsWithoutUnicodeChar.dfy
// RUN: %run "%s" --unicode-char:true --input ../Unicode/UnicodeStringsWithUnicodeChar.dfy
```

So I try
```
git clone https://github.com/dafny-lang/libraries
cd libraries/src/JSON
dafny run --target cs --unicode-char:false --input ../Unicode/UnicodeStringsWithoutUnicodeChar.dfy  Tests.dfy 
dafny run --target js --unicode-char:false --input ../Unicode/UnicodeStringsWithoutUnicodeChar.dfy  Tests.dfy 
dafny run --target java --unicode-char:false --input ../Unicode/UnicodeStringsWithoutUnicodeChar.dfy  Tests.dfy 
```

My results are
* C# fails with "The type or namespace name 'Abs' does not exist in the namespace 'Math' (are you missing an assembly reference?)"
* JavaScript fails with "Cannot find module 'bignumber.js'"
* Java fails with can't find Math.__default

markrtuttle:
Tests.dfy includes API.dfy which includes Serializer.dfy which includes ../Math.dfy which defines Abs.  So verification works but compilation fails.

markrtuttle:
I think the compiler is picking up the standard Math class and not the Math class defined in this library.  Renaming Math to DMath in this library solves the problem for Java, and I think also for C# and JavaScript.

PR #146, opened 2023-10-04
Rename module Math->MathLib, name conflicts with Java/C# Math class

markrtuttle:

This renames the Math module to MathLib to avoid conflict with the Math class in Java and C#.  Prior to this change, the examples in JSON will verify but not compile to Java or C# (error messages about missing symbols indicate the compiler is looking in the standard Math library and not the Dafny Math module).

This does not update the corresponding files under examples.  Should it?


robin-aws (commented):
Thanks for reporting this issue and the proposed fix. I'm not yet sure I want to merge it, though, because it will be a breaking change for anyone using the module, and more importantly hides the underlying issue: code compiled from Dafny should not be dumped into the top-level namespace, where these kinds of conflicts could easily occur for other names in the future as well.

I'd also like to get to the bottom of why the CI didn't catch #145 before we resolve it.

Fortunately we're already working on supporting configuring the parent namespace for compiled code: https://github.com/dafny-lang/dafny/issues/4322. We should definitely use that feature when building library code like this to avoid these problems in general.

markrtuttle:
@robin-aws I've pushed an update that merely moves Math into Dafny.Math.

PR #147, opened 2023-10-04
Replace deprecated syntax 'forall ensures' -> 'assert ... by'

markrtuttle:

This replaces syntax flagged by Dafny 4.2 compiler as deprecated: forall ensures with no quantifiers.


robin-aws (approved):
Thanks for cleaning this up!

PR #149, opened 2023-10-24
Chore: Proofs not depending on leaks of opaque + brittleness reduction

MikaelMayer:
I added reveal statements where they were previously "leaked" because of fuel encoding that are soon going away. I also reduced the RU for ToLarge from 6.8M to 1.3M.


src/Collections/Sequences/LittleEndianNatConversions.dfy
```diff
@@ -81,15 +81,35 @@ abstract module {:options "-functionSyntax:4"} LittleEndianNatConversions {
...
+        LemmaModSubMultiplesVanishAuto();
+      }
+      var ys := ([Small.ToNatRight(xs[..E()]) as Large.uint] + ToLarge(xs[E()..]));
+      assert |ToLarge(xs[E()..])| == |xs[E()..]| / E();
+      assert |ys| == |xs| / E() by {
+        calc {
+          |ys|;
+          1 + |xs[E()..]| / E();
+          1 + (|xs| - E()) / E();
+          { DivMod.ModINL.LemmaFundamentalDivMod(|xs|, E()); }
```
keyboardDrummer, line 104: I don't understand why this rewrite is so complicated.

Couldn't you do:

```dafny
|ys|;
1 + |xs[E()..]| / E();
1 + (|xs| - E()) / E();
1 + |xs| / E() - E() / E();
1 + |xs| / E() - 1;
|xs| / E();
```

MikaelMayer: Your proof works too. The difference is that with my version it takes 1.35M RU, and yours takes 2.16M RU.
The reason why it's so complicated is that I wrote a detailed proof that ought to reduce the resource count, and then gradually removed elements that decreased the resource count, until I arrived at a minimum.

keyboardDrummer: I think it's good to optimize for resource count, but it should remain easy to understand why the code is the way it is. If special tricks are required to achieve a certain resource count, they should be explained. Currently I can't understand why this proof is the way it is. I would have expected the one with fewer steps to use fewer resources.

MikaelMayer: I don't have a full understanding why some terms reduce the resource count and some don't. Some changes might not be statistically significant (like the last effort), some are surely (like putting the lemma ensuring quantifiers in an if then condition)
The only thing I know is that this proof works in 4.0.0, 4.1.0, 4.2.0, 4.3.0 when before it would not, so I'd say we keep it as such, until one day we have a better understanding about proof elements that can predictably and reliably reduce the resource count.

atomb (approved):
This looks good to me. It is slightly disappointing to have a proof where the choice of a particular structure is hard to explain, but the fact that it works with many Dafny versions makes me inclined to accept it. We may be able to continue to improve it later.

src/Collections/Sequences/LittleEndianNatConversions.dfy
keyboardDrummer: Wouldn't the simple version also work with all those versions? If so I'd prefer that.

Otherwise, could you put a comment in the code to explain how you got to that proof?

MikaelMayer: I added a comment about the process.
Note that this file does not use nonlinearities, so it's a bit puzzling that your proof works on the first place.

keyboardDrummer: > Note that this file does not use nonlinearities, so it's a bit puzzling that your proof works on the first place.

Maybe because it doesn't require interpreting any arithmetic operations? It only uses `X / X == 1` and the distributive law.

src/Collections/Sequences/LittleEndianNatConversions.dfy
```diff
@@ -81,15 +81,38 @@ abstract module {:options "-functionSyntax:4"} LittleEndianNatConversions {
...
       LemmaDivMinusOne(|xs|, E());
-      [Small.ToNatRight(xs[..E()]) as Large.uint] + ToLarge(xs[E()..])
+      assert |xs[E()..]| % E() == 0 by {
+        LemmaModSubMultiplesVanishAuto();
+      }
+      var ys := ([Small.ToNatRight(xs[..E()]) as Large.uint] + ToLarge(xs[E()..]));
+      assert |ToLarge(xs[E()..])| == |xs[E()..]| / E();
+      assert |ys| == |xs| / E() by {
+        // To obtain this proof, Mikael Mayer wrote first a detailed proof
+        // Then he removed everything he could that was decreasing the resource count
```
keyboardDrummer, line 101: > To obtain this proof, Mikael Mayer wrote first a detailed proof

What does that mean? How do I write a detailed proof? Also, I would leave out the name, "To obtain this proof, first write a detailed proof using techniques X, then prune it to reduce the resource count"

keyboardDrummer (dismissed):

src/Collections/Sequences/LittleEndianNatConversions.dfy
MikaelMayer: Oh it's because it's not true in general, with natural integers, that
```
(a + b) / c == a / c + b / c
(a - b) / c == a / c - b / c
```
Proof: Take a == b == 1 and c == 2 for the first (1 == 0), and a == 2, b == 1, c == 2 for the second (0 == 1)

So if you want to make the simplification in any case, proving that b == c is necessary and then you need to invoke the lemma that states how to compute an integer division.

src/Collections/Sequences/LittleEndianNatConversions.dfy
```diff
@@ -81,15 +81,38 @@ abstract module {:options "-functionSyntax:4"} LittleEndianNatConversions {
...
+        // until arriving at at a minimum
```
MikaelMayer, line 102:
```suggestion
        // To obtain a proof like this, first write a detailed proof
        // as much as you can, not assuming anything about non-linear arithmetic (use only lemmas for that)
        // Then remove intermediate computation steps and lemma calls if doing so decrease the resource count
        // until arriving at at a minimum
```

keyboardDrummer (approved):

atomb (approved):

src/Collections/Sequences/LittleEndianNatConversions.dfy
MikaelMayer: `a/y` is not the same as `a(1/y)`. `1/y == 0` when `y > 1`

PR #154, opened 2024-01-03
fix: Add missing type characteristics

RustanLeino:
This PR primarily adds missing `(!new)` type characteristics in function declarations. These had gone undetected, because of a latent [bug](https://github.com/dafny-lang/dafny/issues/4926) that has now been [fixed](https://github.com/dafny-lang/dafny/pull/4928). The bug fix requires these fixes in the libraries.

This PR also cleans up some library files: removing deprecated semi-colons, removing unnecessary parentheses, stylizing order of specification clauses, and removing explicit triggers that are saying the same thing as those triggers that are automatically inferred.

~~The PR also changes `forall ensures ...` statements with no bound variables into `assert ... by` statements. (I think we attempted this change once before, only to find that the change tickled some brittleness issue. Let's see if they verify this time.)~~

Finally, the PR improves some `reads fn` clauses, where `fn` is some function. In one case, such a `reads` clause was probably accidentally (and I changed it to the more specific `reads fn(this)`). In the other cases, the `reads fn` clause had ended up reading more than the enclosing function's precondition could support. The new `reads` clauses are more specific than the previous, so, semantically, they should not affect callers. I'm surprised how these were not detected before (am I perhaps running the wrong version of Dafny when running the `libraries` test suite?). In fact, I also received an error, which I _think_ is due to that a previous version of Dafny didn't assume `requires` clauses when checking `reads` clauses in lambda expressions (or was it perhaps the other way around?). If there's still a problem with my modifications, I hope CI will discover them.

Hint to reviewers: If you want to see each of the simple modifications, I suggest looking at each commit individually.

### DIFF
--- a/src/Collections/Sequences/Seq.dfy
+++ b/src/Collections/Sequences/Seq.dfy
@@ -752,11 +752,11 @@ module {:options "-functionSyntax:4"} Seq {
   }
 
   /* inv(b, xs) ==> inv(FoldLeft(f, b, xs), []). */
-  lemma LemmaInvFoldLeft<A,B>(inv: (B, seq<A>) -> bool,
-                              stp: (B, A, B) -> bool,
-                              f: (B, A) -> B,
-                              b: B,
-                              xs: seq<A>)
+  lemma LemmaInvFoldLeft<A(!new), B(!new)>(inv: (B, seq<A>) -> bool,
+                                           stp: (B, A, B) -> bool,
+                                           f: (B, A) -> B,
+                                           b: B,
+                                           xs: seq<A>)
     requires InvFoldLeft(inv, stp)
     requires forall b, a :: stp(b, a, f(b, a))
     requires inv(b, xs)
@@ -810,11 +810,11 @@ module {:options "-functionSyntax:4"} Seq {
   }
 
   /* inv([], b) ==> inv(xs, FoldRight(f, xs, b)) */
-  lemma LemmaInvFoldRight<A,B>(inv: (seq<A>, B) -> bool,
-                               stp: (A, B, B) -> bool,
-                               f: (A, B) -> B,
-                               b: B,
-                               xs: seq<A>)
+  lemma LemmaInvFoldRight<A(!new), B(!new)>(inv: (seq<A>, B) -> bool,
+                                            stp: (A, B, B) -> bool,
+                                            f: (A, B) -> B,
+                                            b: B,
+                                            xs: seq<A>)
     requires InvFoldRight(inv, stp)
     requires forall a, b :: stp(a, b, f(a, b))
     requires inv([], b)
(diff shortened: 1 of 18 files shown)


src/Collections/Sequences/Seq.dfy
```diff
@@ -653,7 +653,7 @@ module {:options "-functionSyntax:4"} Seq {
   /* Applying a function to a sequence  is distributive over concatenation. That is, concatenating 
      two sequences and then applying Map is the same as applying Map to each sequence separately, 
      and then concatenating the two resulting sequences. */
-  lemma {:opaque} LemmaMapDistributesOverConcat<T,R>(f: (T ~> R), xs: seq<T>, ys: seq<T>)
+  lemma {:opaque} LemmaMapDistributesOverConcat<T, R>(f: (T ~> R), xs: seq<T>, ys: seq<T>)
```
MikaelMayer, line 656: An opaque lemma? :-)

src/dafny/Collections/Arrays.dfy
```diff
@@ -15,7 +15,7 @@ module Dafny.Collections.Arrays {
   import opened Relations
   import opened Seq
 
-  method BinarySearch<T>(a: array<T>, key: T, less: (T, T) -> bool) returns (r: Option<nat>)
+  method BinarySearch<T(!new)>(a: array<T>, key: T, less: (T, T) -> bool) returns (r: Option<nat>)
```
MikaelMayer, line 18: I'm very surprised by this type characteristic. I would have thought it would have been possible to sort an array of arrays by the size of each array for example. Why is this required now? Can we relax it if possible?

RustanLeino: It's because of how the specification was written. The precondition uses `StrictTotalOrdering`, which ends up quantifying over all values of the type. If those ordering predicates were instead written to apply only to an element of given set or sequence, then the `(!new)` would not be needed.

src/Collections/Sequences/Seq.dfy
RustanLeino: Thanks, that makes no sense. I removed them. I also removed them in https://github.com/dafny-lang/dafny/pull/4928.

RustanLeino:
Note, this `libraries` repository was recently deprecated (https://github.com/dafny-lang/libraries/pull/155). Still, I think it makes sense to roll in this PR, since it contains valid changes and is otherwise ready to go.

src/dafny/Collections/Arrays.dfy
MikaelMayer: Bumping on this one.

atomb (approved):
I was just looking at making similar changes, and opened #156, but this PR already does more. I'd like to see it merged, since the `libraries` repo is still being used, even if it's deprecated.

robin-aws (approved):

src/dafny/Collections/Arrays.dfy
MikaelMayer: When I wrote my last message, I hadn't seen yours on the thread. I should have refreshed the page first.

PR #158, opened 2024-08-05
Remove warnings in forall

seebees:
We are still using this.
Need to move over to the DafnyStandardLibrary, but that is not compiled with `utf8-false` yet.


robin-aws (approved):
Just to not be sneaky about it :) we're removing `SetToSeq` because it's the only case of non-determinism in the library (and the pending Dafny 4.8 will fix the scope of `--enforce-determinism` to catch it), and too slow for production code anyway.

atomb (approved):

PR #160, opened 2024-10-03
chore: Remove Dafny trigger warnings

seebees:
This removes Dafny warnings around `forall` statments lacking a `{trigger}`

### DIFF
--- a/src/NonlinearArithmetic/DivMod.dfy
+++ b/src/NonlinearArithmetic/DivMod.dfy
@@ -724,7 +724,7 @@ module {:options "-functionSyntax:4"} DivMod {
     ensures forall x: int, a: int, d: int {:trigger a / d, x * d, x * a}
               :: 0 < x && 0 <= a && 0 < d ==> 0 < x * d  &&  a / d == (x * a) / (x * d)
   {
-    forall x: int, a: int, d: int {:trigger a / d, x * d, x * a} | 0 < x && 0 <= a && 0 < d
+    forall x: int, a: int, d: int {:nowarn} | 0 < x && 0 <= a && 0 < d
       ensures 0 < x * d  &&  a / d == (x * a) / (x * d)
     {
       LemmaDivMultiplesVanishQuotient(x, a, d);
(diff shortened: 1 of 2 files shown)


src/NonlinearArithmetic/DivMod.dfy
```diff
@@ -209,7 +209,7 @@ module {:options "-functionSyntax:4"} DivMod {
     ensures forall a: int, b: int, d: int, R: int {:trigger d * ((a + b) / d) - R, d*(a/d) + d*(b/d)}
               :: 0 < d &&  R == a%d + b%d - (a+b)%d ==> d*((a+b)/d) - R == d*(a/d) + d*(b/d)
   {
-    forall (a: int, b: int, d: int, R: int {:trigger true} | 0< d &&  R == a%d + b%d - (a+b)%d)
+    forall (a: int, b: int, d: int, R: int {:trigger d * ((a + b) / d) - R, d*(a/d) + d*(b/d)} | 0< d &&  R == a%d + b%d - (a+b)%d)
```
keyboardDrummer, line 212: I think you can't use interpreted functions, like any of the arithmetic or logical operators, in a trigger. Sorry that there is no automatic feedback on this, or even good documentation. And to be honest I'm not sure what the right trigger for this forall would be. @MikaelMayer @atomb ?

seebees: You can run Dafny and it will give you some help, but also `{:nowarn}` works in newer versions of Dafny, and that gets me what I want!

robin-aws (approved):

ajewellamz (approved):
LGTM

src/NonlinearArithmetic/DivMod.dfy
MikaelMayer: Please use `{:nowarn}` everywhere then. These triggers are not valid triggers because they map to builtin operators that do not work with pattern matching.

ssomayyajula (approved):

PR #162, opened 2024-11-14
fix: Drop SelectOpt from MutableMap

ShubhamChaturvedi7:

Drops `SelectOpt` from MutableMap: https://t.corp.amazon.com/e502d5d5-8d37-4249-bba6-60fe3545a6ea/communication#899a85f9-9b66-418f-90d0-ee0505811d2e

### DIFF
--- a/examples/MutableMap/MutableMapExamples.dfy
+++ b/examples/MutableMap/MutableMapExamples.dfy
@@ -57,8 +57,6 @@ module {:options "-functionSyntax:4"} MutableMapExamples {
     print m.Select("testkey2"), "\n";
 
     m.Remove("testkey");
-    AssertAndExpect(m.SelectOpt("testkey").None?);
-    AssertAndExpect(m.SelectOpt("testkey2").Some? && m.SelectOpt("testkey2").value == "testvalue2");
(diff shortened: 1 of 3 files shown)


ajewellamz (approved):

robin-aws:
Since the link is not publicly accessible, for the record here's my comment that led us here:

> FWIW the copy of MutableMap we include in the Dafny standard libraries avoids this [not being able to implement the class in Go] by not defining any functions/method in Dafny - it doesn't have the problematic SelectOpt helper function: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std/TargetSpecific/Concurrent-notarget-java-cs-js-go-py.dfy#L70

> Although I know we are intending to replace the dafny-lang/libraries submodule with a dafny-lang/dafny submodule so you can use the maintained standard library code, I also know that's going to take substantial work. So I'd suggest as a short term solution we modify libraries to remove the SelectOpt function, since the MPL isn't using it, so that it's possible to implement that MutableMap class in all target languages. Would that work for you?

robin-aws (approved):

MikaelMayer (approved):
Looks good

PR #168, opened 2025-05-09
fix: correct Unicode handling in JSON  

rishav-karanjit:
fixes 
- Unicode escape sequence padding with leading zeros instead of trailing spaces
- Error condition to be only when str legnth is < 6 but not <= 6


MikaelMayer (approved):
Linked PR that contains the tests https://github.com/dafny-lang/dafny/pull/6228

robin-aws (approved):

PR #170, opened 2025-08-15
chore: update mutable maps 

rishav-karanjit:
updating mutable maps based on: https://github.com/dafny-lang/dafny/pull/5999


src/MutableMap/MutableMap.dfy
```diff
@@ -62,8 +62,13 @@ module {:extern "DafnyLibraries"} {:options "-functionSyntax:4"} DafnyLibraries
...
 
   class {:extern} MutableMap<K(==),V(==)> extends MutableMapTrait<K,V> {
-    constructor {:extern} ()
-      ensures this.content() == map[]
+
+    // Invariant on key-value pairs this map may hold
+    ghost const inv: (K, V) -> bool
+
+    constructor {:extern} (ghost inv: (K, V) -> bool, bytesKeys: bool := false)
+      ensures 0 == |this.content().Items|
```
ssomayyajula: Why not keep the old postcondition?

ssomayyajula:
Because `MutableMap` is the only extern in `libraries`, I suppose it's OK for it to be out of sync with `DafnyStandardLibraries` until we figure out what's going on with https://github.com/dafny-lang/dafny/issues/6333, but I'll let the other reviewer decide.

rishav-karanjit:
> Because `MutableMap` is the only extern in `libraries`, I suppose it's OK for it to be out of sync with `DafnyStandardLibraries` until we figure out what's going on with [dafny-lang/dafny#6333](https://github.com/dafny-lang/dafny/issues/6333), but I'll let the other reviewer decide.

Yes. We (from crypto tools) just update it as we need it.

src/MutableMap/MutableMap.dfy
rishav-karanjit: In MPL, I was not able to prove that map was injective after I create a mutable map. Dafny does not understand when we construct new mutable map, the map is empty.

ssomayyajula: To clarify, it's unable to prove `this.content()` is injective when you have `ensures this.content() == map[]`, but it is able to when you have `ensures |this.content()| == 0` or `|this.content().Items| == 0`?

rishav-karanjit: This (below code) is from the PR that Robin made which does not contain any post condition to say the map is empty. When I have this constructor, I am unable to prove the map I create with `new MutableMap(...);` is a injective map. I did not test a post condition `ensures this.content() == map[]`.
```
constructor (ghost inv: (K, V) -> bool, bytesKeys: bool := false)
      ensures this.inv == inv
```
from: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std/TargetSpecific/Concurrent-notarget-java-cs-js-go-py.dfy#L79-L80

ssomayyajula: Oh I see, the version in `DafnyStandardLibraries` doesn't have `content()` but it exposes a copy of the `map` API. Not constraining `Items()` over there seems like an oversight. Not to be nitpicky, but if the original API in `libraries` had `ensures this.content() == map[]`, maybe we should keep it that way?

rishav-karanjit: Yes I am fine with putting `ensures this.content() == map[]`. I think this will be same as saying constructor only make empty map. But, what is difference between `ensures this.content() == map[]` and `0 == |this.content().Items|`?

rishav-karanjit: I changed the post condition to `ensures this.content() == map[]`.

ssomayyajula:
Since `inv` is now part of `MutableMap`, is there any reason why it's not part of the specs as in https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std/TargetSpecific/Concurrent-notarget-java-cs-js-go-py.dfy? Do you not need it?

MikaelMayer (approved):
Works for me

rishav-karanjit:
> Since `inv` is now part of `MutableMap`, is there any reason why it's not part of the specs as in [dafny-lang/dafny@`master`/Source/DafnyStandardLibraries/src/Std/TargetSpecific/Concurrent-notarget-java-cs-js-go-py.dfy](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std/TargetSpecific/Concurrent-notarget-java-cs-js-go-py.dfy?rgh-link-date=2025-08-19T20%3A54%3A03Z)? Do you not need it?

Its not part of spec in this PR because I did not needed that in MPL. You can add a suggestion if you recommend it or add it in a different PR. I am fine with either.

ssomayyajula:
Oh got it, since it's `ghost` I don't have a problem. Approved

ssomayyajula (approved):

